PR #26, opened 2019-05-09
Scenario

### LINKED ISSUE #15: Create a `Scenario` object
Create an object to read and hold information about the scenario, including:
- initialization of scenario with a base network and a 'tag' (e.g. `example` is what is used in [the simple roadway attribute change project card ](https://github.com/wsp-sag/network_wrangler/blob/master/example/stpaul/project_cards/1_simple_roadway_attribute_change.yml). 
- reading and validating project cards from a folder that contain that tag.  
- should continue to read all project cards even if one fails...and then list all the validation fails.

Example API from the PRD document:
`network = create_network(base_network = year_2018, 
                         tags = ["thrive_msp_2040"])`

chryssac:
this fixes #15 

Here is an initial implementation of the Scenario object. I have a couple of questions in the code.


network_wrangler/ProjectCard.py
```diff
@@ -7,7 +7,7 @@
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 from jsonschema.exceptions import SchemaError
-from .Logger import WranglerLogger
+from Logger import WranglerLogger
```
e-lo: This doesn't work for me in the docker dev environment. Does it work for you in the dev environment? 

What I get:

```bash
docker build wrangler-dev
docker run wrangler-dev /bin/bash
docker-compose run wrangler-dev python -c "import network_wrangler"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/jovyan/work/network_wrangler/__init__.py", line 7, in <module>
    from .ProjectCard import ProjectCard
  File "/home/jovyan/work/network_wrangler/ProjectCard.py", line 10, in <module>
    from Logger import WranglerLogger
ModuleNotFoundError: No module named 'Logger'
```

Changing this lint to `from .Logger` fixes it.

PR #36, opened 2019-06-05
Roadway change

### LINKED ISSUE #14: Apply simple roadway feature change
Change the network object according to the [`1_simple_roadway_attribute_change`](https://github.com/wsp-sag/network_wrangler/blob/master/example/stpaul/project_cards/1_simple_roadway_attribute_change.yml) project card.

Should be able to be tested with something like: `roadway_network_object.apply_simple_network_change(file='1_simple_roadway_attribute_change')`

chryssac:
fixes #14 

checked in an initial implementation of the method


network_wrangler/RoadwayNetwork.py
```diff
@@ -116,3 +117,42 @@ def write(self, filename: str, path: str = '.') -> None:
 
         shapes_file = os.path.join(path, filename + "_shape.geojson")
         self.shapes_df.to_file(shapes_file, driver='GeoJSON')
+
+
+
+    def apply_roadway_feauture_change(self, project_card: str) -> bool:
```
e-lo: feauture-->feature [small typo]

network_wrangler/RoadwayNetwork.py
```diff
@@ -116,3 +117,42 @@ def write(self, filename: str, path: str = '.') -> None:
...
+        Changes the road network according to the project card passed
+        
+        args:
+        project_card: the path to the project card
+        
+        returns:
+        bool: True if successful.
+        '''
+        #import pdb; pdb.set_trace()
+        with open(project_card, 'r') as card:
```
e-lo: Ideally the project card object would hold the project card info rather than reading it in again with RoadwayNetwork.

network_wrangler/RoadwayNetwork.py
```diff
@@ -116,3 +117,42 @@ def write(self, filename: str, path: str = '.') -> None:
...
+            card_dict = yaml.safe_load(card)
+            
+            road_dict = card_dict.get("Road")            
+            name_str = road_dict.get(list(road_dict)[0])
+            id = name_str.split('=')[1]
+            
+            change_dict = card_dict.get("Change")
+            existing_value = change_dict.get(list(change_dict)[0])
```
e-lo: existing value should be checked with the value in the network; 
optionally, a failed match would trigger a warning or full fail

network_wrangler/RoadwayNetwork.py
kulshresthaa: so, you would want project card to be read in a separate method and to pass the project card info dict as an arg into the method `apply_roadway_feauture_change`?

kulshresthaa: Never mind! I see it. We already have a class (`ProjectCard.py`) to read project cards. So, it makes sense to pass the project card object to the method `apply_roadway_feauture_change`.

kulshresthaa:
@e-lo question - a single project card will always have information about one single link for roadway change? Also, it will update only one `attribute` for that link?

DavidOry:
> @e-lo question - a single project card will always have information about one single link for roadway change? Also, it will update only one `attribute` for that link?

No. It will update any number of attributes across any set of links. It should provide enough information for the software to select a set of links and then make changes to a set of attributes for those links. Here's the example from the PRD:

``` {yaml}
Project: Mitchell Road Southbound Widening
Category: Roadway Attribute Change
Tags: 
   - thrive_msp_2040
   - thrive_msp_2040_draft
Dependencies:
	Prerequisite: None
      Corequisite: Mitchell Road Northbound Widening
	Conflicts: Mitchell Road Soundbound Bike Lanes
Road:
   Name: Mitchell Road
   A_Intersection: Tamarack Trail
   B_Intersection: Boulder Point Road 
   Intermediate_Intersections:
      - Village Woods Drive
	- Carmody Drive
	- Sorrel Way
	- Chestnut Drive
Attribute: Lanes
Change:
   Existing: 1
   Build: 2
Notes: Placeholder for notes
```

e-lo:
Right. So you will probably want to create a methods in roadway network along the lines of:
   
    select_roadway_features(self, selection: tuple(str or dict)):
        '''
        Selects roadway features that 

        Example:
           net.select_roadway_links(
              selection = [ {'from': {'osmid': '1234'},       # a match condition for the from node using osm, shared streets, or model node # 
                             'to': {'shstid': '4321'},            # a match for the to-node..
                             'facility': {'name':'Main st'},       #a regex or match for facility condition...could be # of lanes, facility type, etc.
                             }, ...
                            ])
        '''



    change_roadway_features(self, selection: (dict or tuple(str) or , changes: list(dict)):
        '''
        Makes selection of changes to features that are part of the selection.

        net.change_roadway_features(
             selection = (see above)
             changes = [ {'lanes':6}, 
                                  {'ffs': 65},
                                  {'changes+': 'superfreeway'}] . # + sign indicates that we should append the value rather than replace it.
                           )

        '''

kulshresthaa:
@DavidOry How will a project card look like for multiple links and multiple attributes to update?
Sequence of `Road`, `Attribute` and `Change`?

kulshresthaa:
@e-lo 
I have the below project card in example 

```
Project: 6th St E Road Diet
Category: Roadway Attribute Change
Tags:
   - example
Dependencies:
  Prerequisite: None
  Corequisite: None
  Conflicts: None
Road:
   Name: id=driveLink88629 # 6th St E
   A: osmid=187854529 # Jackson St
   B: osmid=187899923 # Robert St N
   Intermediate_Intersections: None
Attribute: Lanes
Change:
   Existing: 2
   Build: 1
Notes: Simple project example to do a road diet on 6th St E in front of MetCouncil
```

For the simple roadway attribute change, what I did in the code was to add a function `apply_roadway_feauture_change` which loops over the network links, identifies the link with id as `driveLink88629` and update the `Attribute` (Lanes) for that link from 2 to 1. 

What is the use of A and B `osmid` in the Road?

kulshresthaa:
@e-lo this is what I did to apply a simple one attribute change to a single link. But if the project card will include any number of links with any number of attributes to update, then this will change completely.

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
...
+        # check if the attribute to be updated exists on the network links
+        # get the current attribute value for the link from the network
+        # check for current attribute value to be same as defined in the project card
+        # update the link attribute value
+
+        # TODO:
+        # change desired logger level for the different checks
+
+        found = False
+        for link in self.links_df['features']:
```
e-lo: Since this is pandas, you can avoid a for loop in general and do selections using the following notation:
`https://stackoverflow.com/questions/17071871/select-rows-from-a-dataframe-based-on-values-in-a-column-in-pandas`

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
         shapes_file = os.path.join(path, filename + "_shape.geojson")
         self.shapes_df.to_file(shapes_file, driver='GeoJSON')
 
-
-
-    def apply_roadway_feauture_change(self, project_card: str) -> bool:
+    def apply_roadway_feauture_change(self, card_dict: dict) -> bool:
```
e-lo: feauture-->feature

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
...
-            print(build_value)
-            
-            
-            #check existing value that it matches the attribute field in self.links_df
-            #TODO: how do i manipulate the data in the dataframe to apply the changes?
-            
-            self.links_df
-            
+
+        road_dict = card_dict.get("Road")
```
e-lo: You can do all of this in one line as:

`selection_type, road_id = card_dict["Road"]["Name"].split("=")`

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
...
+        road_id = road_dict.get("Name").split("=")[1]
+
+        attribute = card_dict.get("Attribute").upper()
```
e-lo: typically you would access a dictionary value by using dict['key'] rather than dict.get()
The dict.get() method is slower, but is useful if you want to provide a default value in case it doesn't exist: i.e. 

    my_dict = {'A':1, 'B':2}
    my_dict.get('C', 'not here')

    # returns 'not here'

e-lo: In general though, you probably don't need to create separate local variables for most of these.

network_wrangler/RoadwayNetwork.py
e-lo: I would suggest creating a separate selection method (see my previous note with example) because you will need to select based on several dimensions and several ID types as well as stringing from A to B without a named street.

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
...
+        # get the current attribute value for the link from the network
+        # check for current attribute value to be same as defined in the project card
+        # update the link attribute value
+
+        # TODO:
+        # change desired logger level for the different checks
+
+        found = False
+        for link in self.links_df['features']:
+            if link['id'] == road_id:
```
e-lo: this selection matching should be based on the ID type that is used, it could be shared streets, osm, model id, etc.

network_wrangler/RoadwayNetwork.py
```diff
@@ -118,41 +118,57 @@ def write(self, filename: str, path: str = '.') -> None:
...
+
+                # if projetc card attribute to be updated is not found on the network link
+                if attribute not in link.keys():
+                    WranglerLogger.error('%s is not an valid network attribute!' % (attribute))
+                    return False
+                else:
+                    attr_value = link[attribute]
+
+                    # if network attribute value is not same as existing value info from project card
+                    # log it but still update the attribute
```
e-lo: ##todo eventually whether it logs/warns vs logs/fails should be determined by some runtime settings.

e-lo:
A node and B node will likely be our MAIN selection method in the future since we will want to be updating a string of links going from A to B

kulshresthaa:
Thanks for the comments. I will work on revising the code based on suggestions. 

I see that the stpaul example network was updated today. After reading the network, a network link (`links_df`) looks like this - 
```
{'id': 'bff76735795c148707c774d7f7a0fa6b', 'geometryId': '13e68dc0fe4477f10b9655ebbd5f8435', 'access':
None, 'area': None, 'bridge': None, 'highway': 'tertiary', 'junction': None, 'key': 0.0,
'OSMlanes': 0.0,'length': 112.639, 'maxspeed': None, 'name': 'East Cayuga Street', 'oneway': 'False',
'osmid': '18218485', 'ref': None, 'service': None, 'tunnel': None, 'u': '493882338', 'v': '493882339',
'width': None, 'LINK_ID': 280, 'LANES': 2, 'isTranLink': 0, 'isDriveLink': 1, 'isWalkLink': 0, 'isBikeLink': 0,
'locationReferences': [{'sequence': 1, 'point': [-93.0855338, 44.9662078], 'distanceToNextRef':
112.8086679896,'bearing': 0, 'intersectionId': 'b6ee5e5a4adca2f379b20fe7ee7ca77d'},
{'sequence': 2, 'point': [-93.0843092, 44.9656997], 'intersectionId': '51013ccb2e52ee1ea83de6d36e4ef268'}]}
```

1. Example project card (`1_simple_roadway_attribute_change`) has `id` as selection type and I don't see `id` attribute as that type (`driveLink88629`) in the network. Which attribute in the network will correspond to `id=driveLink88629`?
2. There are no A node and B node attribute on the link. How can we select links based on A/B node? 
Does the example project card needs to be update based on new network or I am not understanding the network/project_card correctly?

network_wrangler/RoadwayNetwork.py
kulshresthaa: > Since this is pandas, you can avoid a for loop in general and do selections using the following notation:
> `https://stackoverflow.com/questions/17071871/select-rows-from-a-dataframe-based-on-values-in-a-column-in-pandas`

The dataframe in links_df is two columns - `type` and `features`. links_df['features'] is a list of dictionaries, I don't think we can apply selection using the pandas select method. 
Instead of for loop, I can do something like below to create a selection

`next(link for link in net.links_df['features'] if link['osmid'] == '18218485')`
OR
`list(filter(lambda link: link['osmid'] == '18218485', net.links_df['features']))`
`

e-lo:
1. there are a few different link IDs: 
- 'id': 'bff76735795c148707c774d7f7a0fa6b'
- 'LINK_ID': 280
- 'osmid': '18218485'
We should be able to select based on any of these based on what is on the left-side of the link, i.e. if Road['name'] =osmid: 1234 then we should query based on the osmid field.

2. A and B can either be selected by 'locationReferences' (1 being A, 2, being B) which then references nodes.geojson or from the geometry ID, which references a shape.geojson record.

`nodes.geojson`
```json
{
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -93.0855338,
                    44.96620780000001
                ]
            },
            "properties": {
                "id": "b6ee5e5a4adca2f379b20fe7ee7ca77d",
                "osmNodeId": "493882338",
                "travelModelId": 3261,
                "isTranNode": 0,
                "isDriveNode": 1,
                "isWalkNode": 1,
                "isBikeNode": 1,
                "outboundReferenceId": [
                    "bff76735795c148707c774d7f7a0fa6b",
                    "42b68a489b91dd5a1415fb25fb53de65",
                    "1846c721da28072c7041d5d9267b7e11",
                    ""
                ],
                "inboundReferenceId": [
                    "e6502f30a2a81ffff6b306f6f5485392",
                    "4abfa037e47c8bf90533d1212885e834",
                    "c632cdd0d44671dca6333eb3d7cd86f9",
                    "",
                    "2f421ae164bcf74f1c4074007c3166d6"
                ]
            }
        },
```
`shape.geojson`
```json
{
            "type": "Feature",
            "geometry": {
                "type": "LineString",
                "coordinates": [
                    [
                        -93.0855338,
                        44.96620780000001
                    ],
                    [
                        -93.0852387,
                        44.966130099999994
                    ],
                    [
                        -93.0850053,
                        44.96605240000001
                    ],
                    [
                        -93.0848096,
                        44.965972199999996
                    ],
                    [
                        -93.0847331,
                        44.96593550000001
                    ],
                    [
                        -93.084592,
                        44.965867899999985
                    ],
                    [
                        -93.0844062,
                        44.965765700000006
                    ],
                    [
                        -93.0843764,
                        44.9657454
                    ],
                    [
                        -93.0843092,
                        44.9656997
                    ]
                ]
            },
            "properties": {
                "id": "13e68dc0fe4477f10b9655ebbd5f8435",
                "shstFromIntersectionId": "b6ee5e5a4adca2f379b20fe7ee7ca77d",
                "shstToIntersectionId": "51013ccb2e52ee1ea83de6d36e4ef268",
                "forwardReferenceId": "bff76735795c148707c774d7f7a0fa6b"
            }
        },
```

network_wrangler/RoadwayNetwork.py
e-lo: Good point. I forgot that it was nested (yay json). Do you think we should normalize it (i.e. https://www.kaggle.com/jboysen/quick-tutorial-flatten-nested-json-in-pandas) or leave it as-is?

kulshresthaa: yes, we could normalize it to make it easy to work with the links dataframe. But if do it in `Read` method and the links_df is flat, would be able to write it out back as nested json in `write` (I assume we want `write` method to output json file)?

e-lo: Yes, you'd also  need to update the existing write method.

kulshresthaa: What I was saying is that if we normalize the link data (in `Read` function) to flat everything (including `locationReferences` nest), we wouldn't be able to write it back as deep nested json object in `write`. What do you think?

kulshresthaa:
@e-lo we need a new example project card that I can use to get going (and test) for roadway change methods.. something for a link with correct link id (`id` or 'LINK_ID` or 'osmid` or A/B node or whatever we plan to use) so that I can select it from the example network!

e-lo:
There you go - i updated the link id. Although the a and b ids are in the u and v columns (we should change that...I'll make note in Sijia's workflow)

e-lo:
OK for basic first implementation.

PR #44, opened 2019-07-17
search and apply method for roadway links

kulshresthaa:
@e-lo 
The two methods seem to work. I haven't done flatten in these for now. Please review if I am moving in the right direction and using the code style as expected. 

TODO: 
1. `select_roadway_features`: build selection query to select multiple links using the info from project card. 
2. `apply_roadway_feature_change`: apply changes for multiple attributes for the selected links

Could you create a temp project card (how it will look like...) with changing multiple attributes, so that I can complete item 2?

Ashish


kulshresthaa:
I am done with these two methods for now, except to account for different format of change specification in project card. Please review when you get a chance!

kulshresthaa:
Made all the changes as discussed. Please review and merge, if looks OK.

PR #45, opened 2019-07-23
check for requisites and conflicts

kulshresthaa:


kulshresthaa:
@e-lo Can you see if something like this makes sense to you?

revised [create_scenario](https://github.com/wsp-sag/network_wrangler/blob/check_for_requisites_and_conflicts/network_wrangler/Scenario.py#L32) to include the requisite and conflicts tags from the project cards

[check_scenario_conflicts](https://github.com/wsp-sag/network_wrangler/blob/check_for_requisites_and_conflicts/network_wrangler/Scenario.py#L87) is checking for conflicts in the project cards for scenario

[here](https://github.com/wsp-sag/network_wrangler/blob/check_for_requisites_and_conflicts/tests/temp_ak.py#L46) is how to create a scenario and call the conflict check.

network_wrangler/Scenario.py
```diff
@@ -38,14 +38,29 @@ def create_scenario(base_scenario: dict, card_directory: str = '', tags: [str] =
         base_scenario: the base scenario
         tags: only project cards with these tags will be read/validated
         folder: the folder location where the project cards will be
+        project_cards_names: project names
         '''
-        scenario = Scenario(base_scenario, project_cards = project_card_list)
 
-        if card_directory:
```
e-lo: see above comment

network_wrangler/Scenario.py
```diff
@@ -38,14 +38,29 @@ def create_scenario(base_scenario: dict, card_directory: str = '', tags: [str] =
...
-            scenario.add_project_cards(card_directory, tags = tags)
+        project_cards = []
 
-        return scenario
+        prereq_dict   = OrderedDict()
+        coreq_dict    = OrderedDict()
+        conflict_dict = OrderedDict()
+
+        for project_name in project_card_names:
```
e-lo: We don't want to have to specify all the card names for all the projects that we might want to apply. We should either be able to:
1. specify a directory and a tag, whereby you end up iterating thru the directory in search of the tag
2. specify a list of project card objects [or dir+names, if you convince me it is better]
3. both (additive)

This function makes it so you are only looking at the named project cards.

network_wrangler/Scenario.py
```diff
@@ -38,14 +38,29 @@ def create_scenario(base_scenario: dict, card_directory: str = '', tags: [str] =
...
+            card = ProjectCard.read(os.path.join(card_directory, project_name + '.yml'))
```
e-lo: this thru line 56 should probably be part of a generic function "add project card" so that it can be called after the scenario is already created.

network_wrangler/Scenario.py
```diff
@@ -69,5 +84,26 @@ def add_project_cards(self, folder: str, tags: [str] = []) -> [ProjectCard]:
...
+        Fail if the project A specifies that project B is a conflict and project B is included in the scenario
+
+        Returns: boolean
+        '''
+
+        conflict_dict = self.conflicts
+        scenario_projects = list(conflict_dict.keys())
+
+        errorFound = False
+        for project, conflicts in conflict_dict.items():
```
e-lo: does this work?

```python
errorsFound = ['ERROR' for p in scenario_projects if conflict_dict[p] in scenario_projects] #but with a better error raise message 

```

network_wrangler/Scenario.py
kulshresthaa: I do agree, there needs to be a function to be able to "add project card(s)" after the scenario is already created. Will work on how that function will be designed after I get the rest of the skeleton for this correct.

network_wrangler/Scenario.py
kulshresthaa: yeah, I saw that you had a function to bulk apply projects cards (with a tag) from a directory.

Co-requisite and Conflicts are not a problem but prerequisite is (based on my understanding so far). For prerequisite, order of how projects cards are applied matters. 

If the projects are read from a directory (looping over .yml files), how do we know in what order they are applied to the scenario? Prerequisite needs the order, correct?

kulshresthaa: The reason I am carrying the list of project card **names** so that in the event of conflicts or missing requisite, you can report which project card (name) has the conflict project and such. If we just have the project card object, we can only report that there is a conflict issue and user will then need to look into the list of all project cards that are applied, to find the issues.

We could have a project card object attribute with `self.name` to simplify things perhaps?

e-lo: I think we need to separate out the queuing, ordering/checking pre-requisites, and then applying as separate steps. Make sense?

network_wrangler/Scenario.py
e-lo: The add project cards method can be used for both the initial add as well as any subsequent one.

network_wrangler/Scenario.py
e-lo: Ooo, good point on the reporting.  If we have a project card object, it should the name within it though...right?

network_wrangler/ProjectCard.py
```diff
@@ -134,3 +134,40 @@ def parallel_managed_lanes(self, card: dict):
         card: the project card stored in a dictionary
         '''
         WranglerLogger.info(card.get('Category'))
+
+    def get_tags(self):
```
e-lo: Probably don't need this method, can be accessed by self.tags

network_wrangler/ProjectCard.py
```diff
@@ -134,3 +134,40 @@ def parallel_managed_lanes(self, card: dict):
...
+        '''
+        Get the project card tag.
+
+        '''
+
+        card_dict = self.__dict__
+        return card_dict['tags']
+
+    def get_dependency(self, dependency_name: str):
```
e-lo: ditto

network_wrangler/ProjectCard.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import os
+import os, sys
 import yaml
 import json
```
e-lo: When read project card in, add a variable for the file it was read from

network_wrangler/Scenario.py
```diff
@@ -2,34 +2,34 @@
...
+from collections import OrderedDict
+from .Logger import WranglerLogger
 
 class Scenario(object):
     '''
     Holds information about a scenario
     '''
 
-    def __init__(self, base_scenario: dict, project_cards: [ProjectCard] = None):
+    def __init__(self, base_scenario: dict, project_cards: [ProjectCard] = None, prerequisite_dict: OrderedDict = None, corequisite_dict: OrderedDict = None, conflict_dict: OrderedDict = None):
```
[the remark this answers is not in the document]
kulshresthaa: forgot to discuss this one. What do you mean having requisites/conflicts dicts contained within base_scenario dict?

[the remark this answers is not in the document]
e-lo: It should be accessible from `base_scenario.prerequisites`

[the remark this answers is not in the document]
kulshresthaa: but those (requisites and conflicts) are not base_scenario attributes. They are for the project cards to be applied for creating a new scenario. I don't understand why they should be within base_scenario.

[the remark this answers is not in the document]
e-lo: Ah. in that case, they should be initiated from within the new Scenario object. I'm not sure what you would be feeding in separately?

kulshresthaa:
@e-lo updated the create scenario process as we discussed. Let me know if the structure/design looks OK now or if I missed something. Also, added functions for conflicts and corequisite checking for your review.

Once confirmed, I will go ahead and work on prerequisite checking.

network_wrangler/Scenario.py
```diff
@@ -2,72 +2,129 @@
...
         folder: the folder location where the project cards will be
+        project_cards_list: list of project cards to be applied
         '''
-        scenario = Scenario(base_scenario, project_cards = project_card_list)
+
+        prereq_dict   = {}
+        coreq_dict    = {}
+        conflict_dict = {}
+
+        for project_card in project_cards_list:
```
e-lo: I think this part happens in add_project_cards()

network_wrangler/Scenario.py
```diff
@@ -2,72 +2,129 @@
...
+        project_cards: this scenario's project cards
         '''
 
         self.base_scenario = base_scenario
         self.project_cards = project_cards
 
-        self.prerequisites = {} # dictionary of
-        self.corequisites  = {}
-        self.conflicts     = {}
+        self.prerequisites = prerequisite_dict
```
e-lo: I think this is initiated as an empty dict and then when projects are added, these are added

network_wrangler/Scenario.py
e-lo: And I think add_project_cards() is called in __init__()

kulshresthaa: `add_project_cards` is to bulk load the projects from a directory to a scenario object (self). I don't think this will be called from init(). But I have added part of its functionality to the init() too. Take a look at the latest commit and let me know.

kulshresthaa:
@e-lo I am done with all the changes for this PR. Please review and let me know if you have comments.

example/stpaul/project_cards/6_test_project_card.yml
```diff
@@ -0,0 +1,21 @@
+name: 6_test_project_card
```
e-lo: I don't think we need to add this here. why not change the value of "project" to "name"? We can save the path where the project card was read from as the filename...

kulshresthaa: yeah, that works too!

kulshresthaa: I'll update the code to reflect this change.

Issue #39, opened 2019-07-08
Apply Multiple Types of Changes from Same Project Card


DavidOry:
Change multiple attributes for same link

e-lo:
Brainstorm of example usage...

```
change_roadway_features(self, selection: (dict or tuple(str) or , changes: list(dict)):
    '''
    Makes selection of changes to features that are part of the selection.

    net.change_roadway_features(
         selection = (see above)
         changes = [ {'lanes':6}, 
                              {'ffs': 65},
                              {'changes+': 'superfreeway'}] . # + sign indicates that we should append the value rather than replace it.
                       )

    '''
```

Issue #38, opened 2019-07-08
Search Method for Affected Roadway Links


DavidOry:
@e-lo to elaborate.

DavidOry:
Initially, it's a simple method but the idea is it returns a set and the action is taken on the set. We can refine the intelligence of how to fetch the vector.

e-lo:
Idea on usage

```
select_roadway_features(self, selection: tuple(str or dict)):
    '''
    Selects roadway features that 

    Example usage:
       net.select_roadway_links(
          selection = [ {'from': {'osmid': '1234'},  # a match condition for the from node using osm, shared streets, or model node # 
                         'to': {'shstid': '4321'},   # a match for the to-node..
                         'facility': {'name':'Main St'},       #a regex or match for facility condition...could be # of lanes, facility type, etc.
                         }, ...
                        ])
    '''
```

e-lo:
put a stub method on this branch: https://github.com/wsp-sag/network_wrangler/tree/change_multiple_links_attributes

e-lo:
The initial pass is working in PR https://github.com/wsp-sag/network_wrangler/pull/47 but doesn't satisfy all use cases and needs to be rigorously tested. Will be creating a notebook to be able to visualize the results.

e-lo:
@kulshresthaa here is a very very rough pseudocode for what we discussed this PM:
https://github.com/wsp-sag/network_wrangler/wiki/Network-search-method

e-lo:
**Question: What should the qualifications be for returning a "good selection"?**
This is showing a path from A-->B that goes along the one-way pair for the specified street because 6th St is one-way the other way.  This seems excessive. 

Thoughts?  cc: @DavidOry @josiekre @kulshresthaa 

![image](https://user-images.githubusercontent.com/794787/62971938-d3bc8600-bdc7-11e9-9c38-a90e4c40f224.png)

e-lo:
Tests pass, calling this good for now.
Opened a separate issue for the criteria for a failing selection.

PR #47, opened 2019-07-31
Change multiple links attributes

e-lo:
Should satisfy and test:
- https://github.com/wsp-sag/network_wrangler/issues/38
- https://github.com/wsp-sag/network_wrangler/issues/39


tests/test_roadway.py
```diff
@@ -2,31 +2,165 @@
...
+    net.write(filename=out_prefix,path=out_path)
+    net_2 = RoadwayNetwork.read(link_file= out_link_file, node_file=out_node_file, shape_file=out_shape_file)
+    print("--Finished:",request.node.name)
+
+@pytest.mark.menow
+def test_select_roadway_features(request):
+    print("\n--Starting:",request.node.name)
+    net = RoadwayNetwork.read(link_file= STPAUL_LINK_FILE, node_file=STPAUL_NODE_FILE, shape_file=STPAUL_SHAPE_FILE, fast=True)
+
+    test_selections = { \
```
e-lo: @kulshresthaa 
We should add a bunch of these to satisfy  and make sure they are returning the right links. 
Right now the last one doesn't pass but we should add others and also make sure they are returning the right links.
I will probably set up a notebook to look at this tomorrow if you don't get to it first.

kulshresthaa: Yes, the last one (5) fails because the selection query builder is only looking at the `link` attributes from the facility dict and not doing anything with `A` and `B`. 

In fact, test selection 1-4 are also not using A and B for selection query. 1 and 2 will return same set of links

I should revise the code to include A and B into selection criteria.

Tell me again, how A and B are used. I mean I know the selection has to return links b/w A and B nodes but how to go about it from links_df and nodes_df?

e-lo: > ...how to go about it from links_df and nodes_df

I don't have a brilliant idea, but I would probably investigate using something like OSMNX

kulshresthaa:
@e-lo Just checking if you are done with the network search changes? Also, anything else left for me to finish for Sprint 4?

kulshresthaa:
**Approach for attribute changes**
```
LANES: 
	existing: 3
	set: 2

LANES:
	existing: 3
	change: -1
```

kulshresthaa:
@e-lo How do you use https://jsonschema.net/ to update the existing project card schema? 
This is for changing project card to specify attribute change.

e-lo:
The tests pass locally, so I am going to merge and then figure out what the heck I'm doing wrong with packaging them for Travis' consumption.

PR #48, opened 2019-08-06
network search implementation and visualization

kulshresthaa:
@e-lo  I've setup the notebook for network search method. Let's discuss (and finalize) the search approach again when you have time.


e-lo:
This actually looks pretty good to me, but I don't think we need to add the non-candidate links. It might be that every iteration we add some big weight to the SP, so 
`link_to_add['weight'] = 100*i`

e-lo:
I already merged this in b/c I don't think it needs to be on separate branches and am updating with some small edits.

kulshresthaa:
I think the reason we need non-candidate links (i.e., entire network) is because it is possible that the A and B nodes aren't connected in the candidate links. 

For this selection: 
```
selection = {
    'link':[
        {'name': ['6th', 'Sixth', 'sixth']}, #find streets that have one of the various forms of 6th
        {'lanes': [1, 2]} # select street segments that have 1 or 2 lanes currently
        ],
     'A':{'osmnodeid': '187899923'}, # start searching for segments at A
     'B':{'osmnodeid': '187942339'}, # end at B
}
```

graph with candidate_links looks like this 
image

which will throw error in finding shortest path 
`sp_route = nx.shortest_path(G, A_id, B_id, weight = 'weight')`

increasing search breadth wont change anything because these set of candidate_links are first set of candidate_links (satisfying A_id and B_id in u and v of links returning from selection[link] query), i.e., its not at all doing `add_breadth` for this. 

```
candidate_links = net.links_df.query(sel_query, engine='python')
node_list_osmnodeids = list(candidate_links['u']) + list(candidate_links['v'])
A_id in node_list_osmnodeids #TRUE
B_id in node_list_osmnodeids #TRUE
```

e-lo:
So we are now correctly doing the following:
- adding breadth to a max_breadth parameter
- calculating weight and creating an SP based on it
- finding the longer route 

But there are still a few hiccups:
1.  we are searching for a route on the wrong way on a one-way street. 
  - Should we add a flag to the method to allow both directions?  
  - Should we return the opposite direction if we select it to be directional?
2. The route found has a detour that might not be desirable. 
  - Should we even return a "clean" selection based on this route? We might prefer that the user be more piece-meal?
image

PR #68, opened 2019-08-20
add managed lane to roadway network

kulshresthaa:


kulshresthaa:
I prefer Option 2 over Option 3 - easier (for me) to read and understand what the project card specify but not sure which one (2 or 3) is better when ML project card starts to get complicated.

e-lo:
Let's think **option 2** over, then for a more complicated example. How about one where it is  hot2 AM peak.

```
LANES:
  - time: [ 7:00, 9:00 ] 
      - set: 2
  - set: 3 . #lack of specification will indicate all the time, or all the rest of the time

ML_LANES:
  - time: [7:00, 9:00]
    - set: 1
  - set: 0

ML_restrictions:
  - time: [17:00, 19:00]
    - hov:
        - set: '2'

ML_price:
  - time: [7:00, 9:00]
    - hov:
        - 1
          - set: '3.25'
        - 2
          - set: '0'

```

does that make sense at all?  [ real question ]

kulshresthaa:
what is
```
ML_restrictions:
  - time: [17:00, 19:00]
    - hov:
        - set: '2'
``` 
in plain English?

e-lo:
Hah! Sorry.. 

> "set the managed lane restriction to be HOV 2"

But then I realized that for a HOT lane you don't want to do that, so really it should be:

```
LANES:
  - time: [ 7:00, 9:00 ] 
      - set: 2
  - set: 3 . #lack of specification will indicate all the time, or all the rest of the time

ML_LANES:
  - time: [7:00, 9:00]
    - set: 1
  - set: 0

ML_restrictions:
 - `none`

ML_price:
  - time: [7:00, 9:00]
    - hov:
        - 1
          - set: '3.25'
        - 2
          - set: '0'
```

kulshresthaa:
Yeah, this works!

`ML_restrictions` need not be part of `properties` dict? Rest of those fields (LANES, ML_LANES, ML_PRICE) will go as is on the network. 

We could have `restrictions` as separate dict in the project card, which could also include information about `connectors` (access/egress) at some point. 

```
properties:
  LANES:
    - time: [7:00, 9:00]
      - set: 2
    - set: 3
  ML_LANES:
    - time: [7:00, 9:00]
      - set: 1
    - set: 0
  ML_PRICE:
    - time: [7:00, 9:00]
      - set:
        - sov: 3.25
        - hov2: 0
    - set: 0
restrictions:
  access:
    trucks: 0
  connectors: 'all'
```

we should finalize some design of project card for a simple ML project and move forward.

e-lo:
Oh wait...I think I meant for restrictions to be part of
properties...otherwise it won't get into the geojson right

kulshresthaa:
What do you mean it wont get into geojson right. Why do we need `restrictions` as part of `properties`? 

`properties` has "attributes" of the links that we will update "as is" when applying the project card. 

`restrictions` would need to be first converted to some type of link attributes before adding them to the links as columns.

e-lo:
Is there a reason why restrictions and a parallel ML_restrictions couldn't go in as a link-level attribute as-is? 

What type of conversion were you thinking of making?

kulshresthaa:
```
  ML_LANES:
    - time: [7:00, 9:00]
      - set: 1
    - set: 0
```
What I am thinking is that the logic for these attributes will be something like - 
if `time` exists - set the attribute `ML_LANES` to 1 (with time information)
else set the attribute `ML_LANES` to 1

Most of the attributes in the `properties` will have this nest structure. If `ML_Restrictions` is inside `properties` but with different nesting structure, it will need to be processed differently.

We aren't creating a field on the link as `ML_Restriction` and the conversion I was thinking was, e.g., create a field `access_truck` or `trucks` with 0 value for
```
restrictions:
  access:
    trucks: 0
```

well, thinking through this, if `ML_Restriction` is inside properties, we will be creating a field `ML_Restriction` and then we could have its values as a nested dictionary but wouldn't that make it complex?

I don't know what I am thinking..

e-lo:
>  If ML_Restrictions is inside properties but with different nesting structure, it will need to be processed differently.

Or we could have it be the same/similar.  At least, that was what I was thinking in my **option 2** proposal:

```
ML_restrictions:
  - time: [17:00, 19:00]
    - hov:
        - set: '2'
```

> ML_Restriction is inside properties, we will be creating a field ML_Restriction and then we could have its values as a nested dictionary

I am thinking that is what we need to do for all of this, but I think that is fine as we are using a json-based format, not csv.

kulshresthaa:
@e-lo @josiekre Could you review if the [example](https://github.com/wsp-sag/network_wrangler/blob/managed_lanes_project/example/stpaul/project_cards/4_simple_managed_lane.yml) managed lane project card looks right?

```
properties:
  - property: LANES
    set: 3
    tod:
      - time: [7:00, 9:00]
      - set: 2
  - property: ML_LANES
    set: 0
    tod:
      - time: [7:00, 9:00]
      - set: 1
  - property: ML_PRICE
    set: 0
    tod:
      - time: [7:00, 9:00]
      - set: 1.5
  - property: ML_RESTRICTIONS
    set: 'None'
```
translates to 

`[{'property': 'LANES', 'set': 3, 'tod': [{'time': [420, 540]}, {'set': 2}]}, {'property': 'ML_LANES', 'set': 0, 'tod': [{'time': [420, 540]}, {'set': 1}]}, {'property': 'ML_PRICE', 'set': 0, 'tod': [{'time': [420, 540]}, {'set': 1.5}]}, {'property': 'ML_RESTRICTIONS', 'set': 'None'}]`

e-lo:
@kulshresthaa : the YML structure for the project card looks good.

I think we'll want to do some special translation as we translate it to the geoDataFrame such that it looks more like:

```
{ "properties": [
     { "LANES": {
         "default": 3, 
          [420, 540]: 2
          },
       "ML_LANES": {
          "default": 0, 
           [420, 540]: 1
          }, 
       "ML_PRICE": {
          "default": 0, 
          [420, 540]: 1.5
          }, 
       "ML_RESTRICTIONS": {
          "default": None
          }
   ]
}

```
Only with the tuples being of datetime objects.


But I just realized/remembered that you can't have an object as a key in JSON so it would be best if it was something slightly different...I'll think.

e-lo:
Ok, thinking this thru, this specification doesn't work for places where price is different for different categories (i.w. HOV vs SOV). 

What about (in GeoDataFrame format this time):

| **LANES** | **ML_LANES** | **ML_PRICE** | **ML_RESTRICTIONS** |
| ------- | ------ | --------- | --------- |
| `{"default": 3, {"time": (17:00, 19:00),"value": 2 }}`     |  `{"default": 0,{"time": (17:00, 19:00),"value": 1 }}`   |  `{"default": 0} `| `{"default": 0,{"time": (17:00, 19:00), "occupancy": 2, "value": 1 }}` |
| `{"default": 3, {"time": (17:00, 19:00),"value": 2 }}`     |  `{"default": 0,{"time": (17:00, 19:00),"value": 1 }}`   |  `{"default": 0,{"occupancy": 2,"value": 1.5 }} `| `{"default": None}` |

The first link entry is a peak HOV2 lane with the second a HOT2

kulshresthaa:
Ahaaa! more complicated :P

So, we will not only change the attribute but also add rows depending on how some of the attributes are specified in project card for managed lane. This will involve a lot of special translation for properties in project card. 

Would it be OK to have duplicated records (with some attributes different but same id's, u, v, location information etc.) in the GeoDataFrame?

kulshresthaa:
Updated project card looks like [this]()
```
properties:
  - property: LANES
    set: 3
    tod:
      - time: [7:00, 9:00]
        set: 2
  - property: ML_LANES
    set: 0
    tod:
      - time: [7:00, 9:00]
        set: 1
  - property: ML_PRICE
    set: 0
    tod:
      - time: [7:00, 9:00]
        set: 1.5
  - property: ML_RESTRICTIONS
    set: 'None'
  - property: ML_ACCESS
    set: 'all'
  - property: ML_EGRESS
    set: 'all'
```
gets read as
`[{'property': 'LANES', 'set': 3, 'tod': [{'time': [420, 540], 'set': 2}]}, {'property': 'ML_LANES', 'set': 0, 'tod': [{'time': [420, 540], 'set': 1}]}, {'property': 'ML_PRICE', 'set': 0, 'tod': [{'time': [420, 540], 'set': 1.5}]}, {'property': 'ML_RESTRICTIONS', 'set': 'None'}, {'property': 'ML_ACCESS', 'set': 'all'}, {'property': 'ML_EGRESS', 'set': 'all'}]`

I translated it to the way you structured above - without the complication of different price for different categories - so no additional row.

Final links_df for selected links looks like - 

image

kulshresthaa:
For properties without `group`

```
  - property: LANES
    set: 3
    timeofday:
      - time: ['7:00', '9:00']
        set: 2
      - time: ['16:00', '19:00']
        set: 2
```
translating (to be stored on network) this as:
`{'default': 3, 'timeofday': [{'time': (25200, 32400), 'value': 2}, {'time': (57600, 68400), 'value': 2}]}`

How do you want to store property with group?

```
  - property: ML_PRICE
    set: 0
    group:
      - category: ['sov']
        timeofday:
          - time: ['7:00', '9:00']
            set: 1.5
          - time: ['16:00', '19:00']
            set: 2.5
      - category: ['hov2']
        timeofday:
          - time: ['7:00', '9:00']
            set: 1.0
          - time: ['16:00', '19:00']
            set: 2.0
```
Option 1: `{'default': 0, 'group': [{category: ['sov'], 'timeofday': [{'time': (25200, 32400), 'value': 1.5}, {'time': (57600, 68400), 'value': 2.5}]}, {category: ['hov2'], 'timeofday': [{'time': (25200, 32400), 'value': 1.0}, {'time': (57600, 68400), 'value': 2}]}]`

Option 2: `{'default': 3, 'timeofday': [{'category': ['sov'], 'time': (25200, 32400), 'value': 2}, {'category': ['sov'], 'time': (57600, 68400), 'value': 2}, {'category': ['hov2'], 'time': (25200, 32400), 'value': 1.5}, {'category': ['hov2'], 'time': (57600, 68400), 'value': 2.0}]}`

@e-lo @josiekre which is better Option 1 or Option 2, or any other ideas?

e-lo:
So basically are you asking if time is nested in category or if category is nested in time?   

I think we want category nested in time.

kulshresthaa:
Yes, that's what I was asking. Option 2 then. 

@e-lo review the code for managed lane changes and if all looks OK, we can call it done.

network_wrangler/RoadwayNetwork.py
```diff
@@ -743,27 +760,26 @@ def apply_roadway_feature_change(self, properties: dict) -> RoadwayNetwork:
...
                         )
                     )
 
-            if "set" in values.keys():
-                build_value = values["set"]
+            if "set" in p.keys():
+                build_value = p["set"]
             else:
-                build_value = values["existing"] + values["change"]
+                build_value = p["existing"] + p["change"]
```
e-lo: question: i think we mean for this to be the existing on the network, not the existing in the project card.

kulshresthaa: yes, if we are just warning (and not throwing error) when the existing in the project card is not same as existing on the network.

e-lo:
I think it is b/c it is querying it in the validation step...

```
ERROR    WranglerLogger:RoadwayNetwork.py:725 ERROR: Properties to change in project card not found in network
ERROR    WranglerLogger:RoadwayNetwork.py:727 ML_LANES specified as attribute to change but not an attribute in network

                                              ML_PRICE specified as attribute to change but not an attribute in network

                                              ML_ACCESS specified as attribute to change but not an attribute in network

                                              ML_EGRESS specified as attribute to change but not an attribute in network
```

network_wrangler/RoadwayNetwork.py
kulshresthaa: done in c4e7fc98ad2a8e01431519c64e50623610308386

e-lo:
@kulshresthaa : Final request before we close the PR.  Please make the `facility` for the project card an actual freeway (bonus if it is part of MnPASS: http://www.dot.state.mn.us/mnpass/mnpassexpresslanes.html)

kulshresthaa:
Can @i-am-sijia identify the managed lane segment in the network file? I don't have a way to identify (visualize) in the json network.

e-lo:
@kulshresthaa : you can visualize it using: http://geojson.io or using folium in a notebook

kulshresthaa:
http://geojson.io throws error when loading link.json file "Invalid JSON file: TypeError: Cannot Convert undefined or null to object".

This is the segment (2 links) in NB direction I added as ML segment for project cards. 

image

Freeway in the stpaul network have empty `name` attribute! Also, `link` in facility dict is required, so the project card has null under 'link' key.

e-lo:
@kulshresthaa I am tied up until later this afternoon and didn't quite get to adding "ref" to the search if something isn't found in "name"...do you want to take a crack at it?  

Essentially if the query doesn't come back with something from 'name', try it again with 'ref' instead.

kulshresthaa:
> @kulshresthaa I am tied up until later this afternoon and didn't quite get to adding "ref" to the search if something isn't found in "name"...do you want to take a crack at it?
> 
> Essentially if the query doesn't come back with something from 'name', try it again with 'ref' instead.

Sorry, I just saw this. I can take a crack at it tomorrow. Let me know!

kulshresthaa:
@e-lo Can you review the change? I replace the "name" with "ref" in the selection query if "name" doesn't return anything in the search method. Test passed.

e-lo:
Looks good to me! 👍👍👍
Merging now.

Issue #63, opened 2019-08-14
Transit route-level change Project Card Consumption 


e-lo:
Update project card read, [project card schema](https://github.com/wsp-sag/network_wrangler/blob/master/schemas/project_card.json), and validation methods to adhere to updates to the project card design.

https://jsonschema.net/ is a useful resource for updating the schema

josiekre:
I think this can now be closed along with #55. Please review @e-lo.

josiekre:
- [ ] Update schema to remove "attribute change" from types of cards

Issue #58, opened 2019-08-14
TransitNetwork route-level attribute change test


e-lo:
Write a test that evaluates if a route-level attribute change was successful.

Issue #57, opened 2019-08-14
TransitNetwork route-level attribute change


e-lo:
Be able to:
- consume a route-level transit change project card
- make changes to one or more routes
- output manipulated TransitNetwork object
- make record of any warnings or errors in the log files

josiekre:
The implementation of `apply_roadway_feature_change()` uses `copy.deepcopy(self)` to create a new network return. I've been attempting to keep as much the same as possible for transit, but it seems like this would create a memory/runtime problem in the case of large networks and many attribute changes. 

https://github.com/wsp-sag/network_wrangler/blob/5edd3c07cc44a30e4c386923eebf2ff77c469fbe/network_wrangler/RoadwayNetwork.py#L725-L745

Why not edit the existing network object at `self`, @e-lo?

e-lo:
I think that is probably fine, but let's make it an option?

josiekre:
> I think that is probably fine, but let's make it an option?

10-4

Issue #56, opened 2019-08-14
Transit Network Object Search Method


e-lo:
A parallel to the search method in RoadwayNetwork.

It should take the information specified in the ProjectCard and then output the relevant components of the transit network.

As a start, let's query by line/route-level features rather than stop/shape-level features (i.e. stop-locations/stop-ids)

PR #73, opened 2019-08-31
Simple transit related issues (sprint 5)

josiekre:
Initiating a pull request into `master` for all transit-related issues in Sprint 5.


josiekre:
I was going to pull in the latest from `master` to make the merge easier, but it is not straightforward. Looks like @e-lo did some work on `master` related to moving schemas around. I did rework the schema for transit (and got it actually failing where it should for roadways).

I'd like to hear from @e-lo on whether you think the three issues (#55, #56, and #63) are closable and then deal with updating this branch to the latest on `master`. Then I'll move to finish up #57 and its test #58.

josiekre:
One thing we need to discuss on Tuesday (cc: @DavidOry) is how we want to treat `start_time` and `end_time` for the first version. I can see at least two common use cases.

## As a selection filter
Find all the trips that start at 06:00:00, end at 09:00:00, and have a `route_id` of 21-111. Then update their headways to 900 secs.

```
facility:
  route_id: 21-111
  times:
    - start: 06:00:00
      end: 09:00:00
properties:
  - property: headway_secs
    set: 900
```

## As a property to change
Find all the trips that have a `route_id` of 21-111. Then update their headways to 900 secs, their `start_time` to 06:00:00, and their `end_time` to 09:00:00.

```
facility:
  route_id: 21-111
properties:
  - property: headway_secs
    set: 900
  - property: start_time
    set: 06:00:00
  - property: end_time
    set: 09:00:00
```

These two result in different networks. 

FYI I've built a weird combination of them right now, where times specified in the properties actually filter the facilities (aka trips) in the selection. When we decide which method to build for first, I will make changes.

josiekre:
We are going to build for using times as a selection filter and not as a property to change. For now we are building as if times on the frequency table are `enums`, where they can only be peak or off-peak. In other words, pretend that they are not numbers.

josiekre:
I am having a hard time with `yaml.safe_load()` and JSON schema being happy with time together.

YAML tries to convert *some* time strings to integers. Here is what I gathered so far:

```
# YAML
time: [9:00, 15:00]

# In Python
{'time': [540, 900]}  # mins from midnight


# YAML
time: [09:00, 15:00]

# In Python
{'time': ['09:00', 900]}  # string, mins from midnight


# YAML
time: [6:00:00, 9:00:00]

# In Python
{'time': [21600, 32400]}  # secs from midnight


# YAML
time: [06:00:00, 09:00:00]

# In Python
{'time': ['06:00:00', '09:00:00']}  # string, string
```

Since the JSON schema validation happens after the `yaml.safe_load()`, we have to write the schema and code to allow all the outcomes if we want users to be able to input unquoted times with or without leading zeros and with or without seconds (a.k.a, as user-friendly as possible).

Worse though, is that if the time comes in as an `int`, we technically do not know if the unit is minutes from midnight or seconds from midnight because `TransitNetwork` cannot see the original project card; it only sees the resulting `int`. (FYI Partridge stores all times as seconds from midnight so we need to get to that eventually.)

In conclusion, I think we need the YAML to always be quoted in the project card so that it definitely comes in as a `string`. (Note that a project card will be able to squeak by validation without the quotes if it uses leading zeros on both times.) In this case we still at least allow users to specify time with or without leading zeros and with or without seconds.

```
time: ['9:00', '15:00']
time: ['09:00', '15:00']
time: ['6:00:00', '9:00:00']
time: ['06:00:00', '09:00:00']
```

Do any of our YAML experts know of a setting or method for otherwise dealing with this surprise `int`? (cc: @e-lo , @DavidOry, @kulshresthaa)

josiekre:
@e-lo I believe #57 and #58 are ready to close now too. Ready for your complete review. I will pause until I hear from you.

e-lo:
I pushed a commit that fixes some more issues with updated schema and passes a few more tests. Outstanding issue from this PR is RoadwayNetwork query validation, which should be addressed in other PR.

PR #166, opened 2020-03-17
147 Add new stop if req for routing change

### LINKED ISSUE #147: Add new stop if required for routing change.
Right now we get a value error if something isn't already listed as a stop and we try and make it one. 

Desired behavior would be to create a new stop.

```python
# If node IDs in properties["set_stops"] are not already
            # in stops.txt, create a new stop_id for them in stops
            if any(
                x not in stops[TransitNetwork.STOPS_FOREIGN_KEY].tolist()
                for x in properties["set_stops"]
            ):
                # FIXME
                WranglerLogger.error(
                    "Node ID is used that does not have an existing stop."
                )
                raise ValueError
```

e-lo:
Should fix #147

Right now we get a value error if something isn't already listed as a stop and we try and make it one.
Desired behavior would be to create a new stop.


josiekre:
The conflict will be with the SCALAR. Looks like you renamed that to SHAPE_ID_SCALAR, but I used the same scalar for both fixes.

network_wrangler/TransitNetwork.py
```diff
@@ -646,15 +648,27 @@ def _apply_transit_feature_change_routing(
...
+                if fk_i not in existing_fk_ids:
+                    WranglerLogger.info(
+                        "Creating a new stop in stops.txt for node ID: {}".format(fk_i)
+                    )
+                    # Add new row to stops
+                    stops.loc[len(stops.index) + 1, [
+                        "stop_id", "stop_lat", "stop_lon",
+                        TransitNetwork.STOPS_FOREIGN_KEY
+                    ]] = [
+                        str(int(fk_i) + TransitNetwork.STOP_ID_SCALAR),
```
e-lo: We would need to verify that whatever we use for `stop_id` is unique (either `fk_i` *or* `fk_i` + a scalar )

e-lo:
>The conflict will be with the SCALAR. Looks like you renamed that to SHAPE_ID_SCALAR, but I used the same scalar for both fixes.

I see that now. Easy for me to change back when this get merged...

network_wrangler/TransitNetwork.py
josiekre: Done.

PR #299, opened 2022-04-22
Allow base scenario as scenario instance

### LINKED ISSUE #298: [FEATURE] Initialize scenario with base scenario as Scenario instance not just dictionary
### User Story
*As a user, I'd like to initialize a scenario with an scenario instance rather than having to access its __dict__ attribute.*

### Priority 
Med

### Level of Effort
Low

### Resolution Ideas
As implemented in "works with ranch" branch: allow either.

#### Tests
Added test: `test_scenario_initialization()`

e-lo:
Closes #298 to allow base scenario be a scenario instance.

Also includes various updates/fixes to typehints/docstrings in `scenario.py` and PEP8 updates.

Tested via: `test_scenario_initialization()`

### DIFF
--- a/network_wrangler/scenario.py
+++ b/network_wrangler/scenario.py
@@ -7,7 +7,7 @@ import glob
 import copy
 import pprint
 from datetime import datetime
-from typing import Union
+from typing import Any, Collection, Mapping, Union
 
 import pandas as pd
 import geopandas as gpd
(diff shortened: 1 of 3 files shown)


DavidOry (changes requested):
1. There are a lot of changes here, only a few related to allowing the `base_scenario` to be a scenario. Can we move the formatting and file moves to a separate PR? (This may not be a good use of our time with the current state of things, but worth asking). 
2. Creating a default value for a new variable also seems to be a separate action. Can that be moved to a separate PR as well? (Same comment about it may not be worth it)
3. I see a test using the `Scenario` constructor, but not one with the static `create_scenario` method. Can we add one with `create_scenario`? This is what Met Council has been using.

e-lo:
Dupe functionality already solved in v1 branch

PR #301, opened 2022-04-23
Create pytest fixtures

### LINKED ISSUE #300: [FEATURE] Use network fixture for testing
### User Story
*As a developer, I'd like tests to run faster by avoiding re-reading in networks*

### Priority 
Medium

### Level of Effort
Low-Medium

### Resolution Ideas
Create a pytest fixture with a factory which can generate copies test networks:
- https://docs.pytest.org/en/6.2.x/tmpdir.html#tmpdir-factory-example
- https://stackoverflow.com/questions/21583833/multiple-copies-of-a-pytest-fixture

#### Tests
- all tests which use highway network

e-lo:
This pull request creates function-level and session-level [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) in order to:

1. Reduce repetition in testing code
2. Increase speed at which testing code completes (a couple vs dozens of minutes)

Speed increase is achieved by caching "expensive" transit network and roadway network reads from disk as a "session-wide" fixture and then deep-copying those instances into the testing functions. 

Note: The code in order to do this is not super streamlined and I'd love for somebody to look it over to see if they can figure out a [successful] way to get this to work without so many sub-fixtures and if/thens.  I have tried dozens of things but this is as good as I could come up with for now.

Closes #300 

cc: @lmz for review (@DavidOry can you add MTC to the dev team?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - pip install -r dev-requirements.txt
   - pip install .
 script:
-  - pytest -m travis
+  - pytest -m -noci
 
 after_success:
   - cd docs; make html
--- a/network_wrangler/conftest.py
+++ b/network_wrangler/conftest.py
@@ -1,0 +1,119 @@
+import os
+from copy import deepcopy
+import logging
+import pytest
+
+from network_wrangler import RoadwayNetwork, ProjectCard, Scenario, TransitNetwork
+
+###############################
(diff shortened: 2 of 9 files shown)


e-lo:
Will update this once #310 is merged, which partially implements this.

e-lo:
Converting this to draft since most of this functionally has already been added to `develop`

PR #311, opened 2023-03-02
Use GH Actions instead of Travis

### LINKED ISSUE #291: [CHORE] Switch from Travis to GH Actions
As a developer, I'd like to use a CI service which more tightly integrates with other things I'd like to do, like publish documentation.

The necessary migration from travis.org --> travis.com provides an opportunity to do so.

e-lo:
This pull requests implements basic functionality to run tests using GH Actions for each push event.

Also:
- Updated some of the basic documentation about running tests and documentation to match
- Separate requirements into different files so don't need to install all potentially messy requirements if you don't need them
- Each requirements file is now an install_extra in setuptools so can be installed when installing network wrangler like pip install .[tests,docs,viz] etc.

Closes #291


requirements.txt
```diff
@@ -6,4 +6,3 @@ partridge >= 1.1
 pyyaml
 geojson
 jsonschema
-folium
```
i-am-sijia, line 9: suggest keeping. there is `import folium` in  "roadwaynetwork.py" and "Roadway Network Search.ipynb"

requirements.viz.txt
```diff
@@ -0,0 +1,3 @@
+folium
+jupyter
+notebook
```
i-am-sijia, line 3: Is this file being used in the GH Actions? Did a global search, did not see any mentioning of this file.

setup.py
```diff
@@ -15,12 +15,14 @@
     long_description = f.read()
 
 with open("requirements.txt") as f:
-    requirements = f.readlines()
+    install_requires = [r.strip() for r in f.readlines()]
 
-with open("dev-requirements.txt") as f:
-    requirements += f.readlines()
+EXTRAS = ["tests"]
```
i-am-sijia, line 20: I think we need to add "docs", and "viz" to `EXTRAS`?

setup.py
```diff
@@ -15,12 +15,14 @@
...
+extras_require = {}
+for e in EXTRAS:
+    with open("requirements.{e}.txt") as f:
```
i-am-sijia, line 23:
```suggestion
    with open(f"requirements.{e}.txt") as f:
```

i-am-sijia (changes requested):

requirements.txt
e-lo: Idea is to separate out all the requirements that are being used vs not...so when GH Actions runs it isn't bogged down and can run efficiently. BUT - it wont pass until I refactor `roadwaynetwork.py` to separate out  viz stuff...which I need to do but was putting off until the other big roadwaynetwork refactor was added. 

But...i can see that it is annoying b/c nothing works w/out it at the moment so I'll add back in.

e-lo:
@i-am-sijia  - the tests are now running in GH Actions (got past flaking) so I think ready for your re-review/approval before I merge when you get a min!

docker/Dockerfile.ci
```diff
@@ -13,7 +13,7 @@ RUN apt-get update \
 COPY requirements.txt /tmp/
 RUN pip install --no-cache-dir --requirement /tmp/requirements.txt
 
-COPY dev-requirements.txt /tmp/
+COPY requirements.dev.txt /tmp/
```
i-am-sijia, line 16:
```suggestion
COPY requirements.tests.txt /tmp/
```
`tests`?

i-am-sijia (approved):

Issue #290, opened 2022-04-19
[FEATURE] Allow kwargs for roadway network creation


e-lo:
### User Story
*As a  user who might use a different field as a foreign ID, I'd like to add kwargs to roadway network creation to override any defaults such as `shape_id` vs `id` for `SHAPE_UNIQUE_ID`.*

### Status

- [x] Defined
- [x] Planned
- [ ] Implemented
- [ ] Tested

### Priority 
Medium

### Level of Effort
Low

e-lo:
WIth new data model implementation we are being more rigorous in what can go in/out of wrangler.  That said, we did implement a small work-around that will rename `id` to `shape_id` in the roadway `shapes.txt`.  This will NOT make it write out that way though!

PR #312, opened 2023-03-02
Allow kwargs in network creation to override Class-level variables for Roadway and Transit

e-lo:
This PR adds steps to the `read()` function for both RoadwayNetwork and TransitNetwork which will read in kwargs from the `read()` variables and will update any matching (upper-case) class-level variables to the specified values.

This is useful for when an agency might need to use a different unique ID or variable set from the defaults. 

Also:
- added util method: `class_vars()`
- test network links which use different Key than `model_link_id`

Tested via: `test_kwarg_instantiation()`

Closes Issue #290

### DIFF
--- a/network_wrangler/roadwaynetwork.py
+++ b/network_wrangler/roadwaynetwork.py
@@ -32,7 +32,7 @@ from .logger import WranglerLogger
 from .projectcard import ProjectCard
 from .utils import point_df_to_geojson, link_df_to_json, parse_time_spans
 from .utils import haversine_distance, create_unique_shape_id, offset_location_reference
-from .utils import create_location_reference_from_nodes, create_line_string
+from .utils import create_location_reference_from_nodes, create_line_string,class_vars
 
 
 class NoPathFound(Exception):
@@ -163,7 +163,7 @@ class RoadwayNetwork(object):
 
     UNIQUE_LINK_KEY = "model_link_id"
     UNIQUE_NODE_KEY = "model_node_id"
-    UNIQUE_MODEL_LINK_IDENTIFIERS = ["model_link_id"]
+    UNIQUE_MODEL_LINK_IDENTIFIERS = [UNIQUE_LINK_KEY]
     UNIQUE_NODE_IDENTIFIERS = ["model_node_id"]
 
     UNIQUE_SHAPE_KEY = "shape_id"
@@ -171,9 +171,9 @@ class RoadwayNetwork(object):
     MANAGED_LANES_REQUIRED_ATTRIBUTES = [
         "A",
         "B",
-        "model_link_id",
         "locationReferences",
-    ]
+    ] + [UNIQUE_LINK_KEY]
+
 
     KEEP_SAME_ATTRIBUTES_ML_AND_GP = [
         "distance",
@@ -236,7 +236,7 @@ class RoadwayNetwork(object):
 
     @staticmethod
     def read(
-        link_file: str, node_file: str, shape_file: str, fast: bool = True
+        link_file: str, node_file: str, shape_file: str, fast: bool = True,**kwargs,
     ) -> RoadwayNetwork:
         """
         Reads a network from the roadway network standard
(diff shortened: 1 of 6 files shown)


tests/test_roadway.py
```diff
@@ -1244,3 +1243,24 @@ def test_managed_lane_restricted_access_egress(request):
...
+@pytest.mark.menow
+def test_kwarg_instantiation(request):
+    WranglerLogger.info(f"--Starting: {request.node.name}")
+
+    update_values = {
+        'unique_link_key':'different_link_id'
+    }
+
+    net = RoadwayNetwork.read(
+        link_file=SMALL_LINK_FILE,
```
i-am-sijia, line 1256: Shouldn't we be testing with the link_diffIDs.json?

network_wrangler/roadwaynetwork.py
```diff
@@ -163,17 +163,17 @@ class RoadwayNetwork(object):
 
     UNIQUE_LINK_KEY = "model_link_id"
     UNIQUE_NODE_KEY = "model_node_id"
-    UNIQUE_MODEL_LINK_IDENTIFIERS = ["model_link_id"]
+    UNIQUE_MODEL_LINK_IDENTIFIERS = [UNIQUE_LINK_KEY]
```
i-am-sijia, line 166: In the example test [`test_kwarg_instantiation()`](https://github.com/wsp-sag/network_wrangler/blob/06740ad41b26418c7ef1276260fc9e59c9b0df0f/tests/test_roadway.py#L1252C8-L1252C8) in test_roadway.py, we update the `UNIQUE_LINK_KEY` to `"different_link_id"`, which works, but it does not automatically update the reference of `UNIQUE_LINK_KEY` in the class, i.e., the `UNIQUE_MODEL_LINK_IDENTIFIEDS` here would still be `["model_link_id"]`, which is causing the test using "links_DiffIDs.json" to fail.

We need to refresh all class attrs whenever setattr() is called.

network_wrangler/transitnetwork.py
```diff
@@ -125,6 +128,19 @@ def read(feed_path: str) -> TransitNetwork:
...
+        """
+        Update Class Variables
+        """
+        _class_vars = class_vars(TransitNetwork)
+        for k,v in kwargs.items():
+            if k.upper() not in _class_vars:
+                WranglerLogger.warning(f"{k} specified in TransitNetwork kwargs but does not match a Class-level variable.")
+                continue
+            WranglerLogger.debug(f"Updating TransityNetwork.{k.upper()} to: {v}")
+            setattr(RoadwayNetwork,k.upper(),v)
```
i-am-sijia, line 140: Here for the same issue in roadwaynetwork.py

network_wrangler/roadwaynetwork.py
```diff
@@ -258,7 +260,18 @@ def read(
...
+        """
+        Update Class Variables
+        """
+        _class_vars = class_vars(RoadwayNetwork)
+        for k,v in kwargs.items():
+            if k.upper() not in _class_vars:
+                WranglerLogger.warning(f"{k} specified in RoadwayNetwork kwargs but does not match a Class-level variable.")
+                continue
+            WranglerLogger.debug(f"Updating RoadwayNetwork.{k.upper()} to: {v}")
+            setattr(RoadwayNetwork,k.upper(),v)
```
i-am-sijia, line 272: after this setattr(), can we refresh all the attrs?

tests/test_roadway.py
i-am-sijia: Test fails when using the link_diffIDs.json. Error:

```python
2023-07-06 15:20:19, ERROR: Network doesn't contain unique link identifier: model_link_id
```

i-am-sijia (changes requested):

e-lo:
Closing this as this approach has been superceeded by aligning on a string model for IDs across all tables in the `selection-refactor` branch.

PR #310, opened 2023-03-01
Fix geometry and indexing of new/modified roadway nodes/links

e-lo:
This PR contains major refactoring of several key pieces of RoadwayNetwork to be more efficient, resilient and understandable.

Improves:
- Efficiency in apply managed lane, add/delete and modify property changes
- Decomposes many monolithic functions and files (still many more to address)
- Adds `num_managed_lane_links()` method to easily test for managed lanes (#294)
- DRY principles
- Efficiency in tests: uses small network when able to, uses fixtures for network so only need to read in once (#300)
- Comprehensiveness in tests: actual assert statements, more tests
- Organization of tests: into smaller files so that can easily run them and understand what each does
- Organization of model-network functions into separate file as a model for decomposing some of roadway network which is...massive
- added documentation at function level
- formatting

Fixes:
- How geometries are modified when nodes/links are added/deleted/modified and for managed lanes (#297)
- How indices are modified when nodes/links are added/deleted/modified and for managed lanes
- A LOT of smaller bugs

...still a lot to do and that this PR doesn't fix.

Issue #287, opened 2022-04-08
[FEATURE] Put logs in their own file


e-lo:
### User Story
*As a model user, I'd like to have my log files live with the output data files and be separated by info and debug rather than having them clutter my base directory.*

### Priority 
Low

### Level of Effort
Low

### Resolution Ideas
Many are already in "changes from MTC" branch.  

Reviewers: @lmz (b/c she is the logging 👑 )

### Risk
Low

#### Tests
General

lmz:
My preference is for the logs to live alongside their output data files rather than in a specific fixed directory.  (e.g. see #286 ) For example, when we build our networks, I like to have the output folder structure look like this:
- network_PBA50_v02
  - 2015 = folder with 2015 networks
  - 2019 = folder with 2019 networks
  - 2022 = folder with 2019 networks
  - 2025 = folder with 2019 networks
  - (etc through horizon year)
  - network_build_20220401_1015_INFO.log = this includes a list of each project applied before each write; this file can be consulted to answer "what projects were included in the year XXXX network", etc.
  - network_build_20220401_1015_DEBUG.log = this includes debug details from the network building process, including validation issues that don't cause a hard failure in the network building process, but modelers might want to be aware of as a possible issue

So the location and name of the log is at the discretion of the script or notebook that created the log, but I tend to think they're most useful when sitting with the output for future forensic purposes.

e-lo:
Addressed in merged PR

PR #313, opened 2023-03-03
Address MTC Logging Concerns

### LINKED ISSUE #286: [FEATURE] setupLogging call in __init__.py is overly prescriptive
### User Story
*As a modeler creating networks, I would prefer to explicitly setup logging rather than automatically in network_wrangler's `__init__.py`*:

https://github.com/wsp-sag/network_wrangler/blob/6ed22407594e4e923f578d56455cb254058933ec/network_wrangler/__init__.py#L40-L46

This is because we find it most useful to have the logs be relevant to what is being done, e.g.
* `build_pipline_network_[date-time].log`
* `apply_projects_[date-time].log`
* `convert_standard_network_to_geopackage_[date-time].log` 
etc

I believe that having this line on the import means that even if we call `setupLogging()` later to create our own log file, we'll now have two, because the import will create one:

```
(tm2_network_dev) PS C:\temp\logtest> python
Python 3.7.9 (default, Aug 31 2020, 17:10:11) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import network_wrangler
>>> network_wrangler.setupLogging(log_filename="custom_log_test.log")
>>> network_wrangler.WranglerLogger.info("This is a test")
2022-03-24 10:30:35, INFO: This is a test
>>> quit()
(tm2_network_dev) PS C:\temp\logtest> ls


    Directory: C:\temp\logtest


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        3/24/2022  10:30 AM             43 custom_log_test.log
-a----        3/24/2022  10:30 AM             43 network_wrangler_2022_03_24__10_30_04.log


(tm2_network_dev) PS C:\temp\logtest> type .\custom_log_test.log
2022-03-24 10:30:35, INFO: This is a test
(tm2_network_dev) PS C:\temp\logtest> type .\network_wrangler_2022_03_24__10_30_04.log
2022-03-24 10:30:35, INFO: This is a test
```

### Priority 
High
### Level of Effort
Low
### Resolution Ideas
I would prefer to remove this line from `__init__.py` and let the user call `setupLogging()`.  
Ideally, we could call this later if the logger is invoked without having being setup -- is that possible?

### Project
*Is there a funder or project associated with this feature?*
MTC
### Who should be involved?
Implementer:  @yuqiww
Commenters: 
Users: 
Reviewers: 

### Risk
I don't think so, especially if we could setup logging later if needed as described above.

#### Tests
We could create a test to verify that both situations work as intended -- if the user sets up a logger explicitly and if the user does not.

e-lo:
This PR implements changes to logging consistent with (what I believe are) MTC/ @lmz desires, including:

1. separate INFO and DEBUG log files
3. DON'T auto-call `setup_logging()`
4. DO auto-call `setup_logging()` when running tests (added a fixture), putting output in tests/out in filenames which will get overwritten so we don't amass gigs of log files.

Also:
- fixed some print statements to be logging instead in some tests

**Note This PR tiers from #310 PR and should be reviewed/merged after that one.**

Fixes #286 #287 

---

RE using a PR tiered from another PR and why we are doing this:
https://softwareengineering.stackexchange.com/questions/432322/git-branching-model-that-support-small-changes-and-downstream-dependencies


lmz (approved):
I didn't review all the changes 'cos there are a lot but the logging updates look good to me.  Thanks for doing this!

PR #314, opened 2023-03-03
Log project card filename

### LINKED ISSUE #274: [FEATURE] Project cards should log their own filename
### User Story
*As a modeler building networks and applying project cards, my logs show things like this:

> 2022-01-13 16:19:08, INFO: Applying Year 2015 Modify Facility Type to Ramp shape incorrect

What project card is that referring to?  I'd rather know what the filename is called than have to search my project cards for that text.  What if multiple cards have the same description?  (or is that a title?).  

I think it would be more useful to say something like:

> 2022-01-13 16:19:08, INFO: Applying [year_2015_roadway_property_change_should_be_ramp_shape_incorrect.yml]: Applying Year 2015 Modify Facility Type to Ramp shape incorrect

### Priority 

To me?  High ;)

### Level of Effort

low

### Project
MTC

### Who should be involved?
Implementer: 
Commenters: 
Users: 
Reviewers: 

### Risk

#### Tests

e-lo:
Log project card filename per MTC request.

Fixes #274

### DIFF
--- a/network_wrangler/scenario.py
+++ b/network_wrangler/scenario.py
@@ -539,7 +539,9 @@ class Scenario(object):
             p = p.__dict__
 
         if p.get("project"):
-            WranglerLogger.info("Applying {}".format(p["project"]))
+             WranglerLogger.info(f"Applying [{p['file']}]:{ p['project']}")
+
+        # for changes, iterate through the changes by calling this method for each change
 
         if p.get("changes"):
             for pc in p["changes"]:


i-am-sijia (approved):

e-lo:
Fixed in: https://github.com/wsp-sag/network_wrangler/pull/327/commits/1f8e26360f2e828af5aa78f6ccc059a84b71f2a2

PR #327, opened 2023-04-13
V1 Refactor RC

e-lo:
This PR contains refactors, bug fixes, performance enhancement, and features for v1 of network_wrangler documented in `notebook/releaseoverviewv1.ipynb` including:

## Features

1. I/O and Mutation Speed
2. Implicit and fast validation
3. Flexible serialization formats
4. Improved stability
5. Selection flexibility for transit based on any feature or by link or node
6. More flexible and fast transition to geodataframes for (almost) any data part
7. Easy data clipping to geographic bounds
8. Scripts for actions you might want to execute from command line (i.e. data conversion/clipping/etc)
9. Logging + Error descriptions
10. Additional project card types: add/delete transit
11. Scripts to create example data from OSM...anywhere

## Tech Overhead Investment

1. Separation of causes prevents circular reference collisions, limits import bloat, and improves legibility/organization.
    - Project card functionality in project card repo
    - Separate modules for separate functionality
2. Reliability achieved through more testable code and expanding test coverage with more test cases and also anti-patterns
    - Testable code
    - Test coverage
3. Explicit data models that make complex data structures obvious and easy to validate
    - Legible
    - Self-documenting
    - Flexible
    - Validatable
4. Clean code principles that make code easier to test and maintain
    - more functions that do a single thing...and do it well
    - classes that are small - functions that access or manipulate them
5. Documentation
    - consistent and detailed functional documentation
    - consistent type hints
    - usage for modules and classes
    - overall documentation leverages the flexible and less bloated MkDocs package
6. Removes less-well-maintained dependencies
    - Replaces Partridge with internal functionality

### Speed

1. Makes the heavy shapes.geojson optional and doesn't read it in for operations that don't directly involve it.
2. Leverages `pandera` for speedy, vector-based data model validation for dataframes
3. Replaces row-based calculations for blank geographic values with vector-based calculations
4. Provides flexibility for I/O serialization formats that are faster – like Parquet.


i-am-sijia:
We used to dump scenarios to pickles via `pickle.dump(scenario, open(scenario_pickle_file, 'wb'))`, and load via `pickle.load(open(scenario_pickle_file, 'rb'))`. Is there an equivalent way to doing this now?

e-lo:
> We used to dump scenarios to pickles via pickle.dump(scenario, open(scenario_pickle_file, 'wb')), and load via pickle.load(open(scenario_pickle_file, 'rb')). Is there an equivalent way to doing this now?

I guess you could do the same thing?  We haven't created a serialized format for scenarios but I think we could obviously think of such a thing as an added feature.

network_wrangler/transit/projects/add_route.py
```diff
@@ -89,71 +89,66 @@ def _add_route_to_feed(
...
+                trip_id = f"trip{j}_shp{shape_id}"
+                stop_dicts = _get_stops_from_routing(trip["routing"])
+                add_trips_df = _create_new_trips(trip, route["route_id"], trip_id, shape_id)
+                add_freqs_df = _create_new_frequencies(headway, trip_id)
+                add_stop_times_df = _create_new_stop_times(stop_dicts, trip_id)
+                add_stops_df = _create_new_stops(
+                    add_stop_times_df["stop_id"], feed.stops["stop_id"], road_net
+                )
+
+                if route['agency_raw_name']:
```
e-lo, line 116: I don't think we need this in our datamodel anymore.  Not sure what functionality it is providing?

network_wrangler/transit/projects/add_route.py
```diff
@@ -89,71 +89,66 @@ def _add_route_to_feed(
     add_routes_df = pd.DataFrame(
         [{k: v for k, v in r.items() if k != "trips"} for r in add_routes]
     )
-    routes_df = pd.concat([feed.routes, add_routes_df], ignore_index=True, sort=False)
+    feed.routes = pd.concat([feed.routes, add_routes_df], ignore_index=True, sort=False)
```
e-lo, line 92: why is feed.routes being set here but all the others are set at the end? We don't want to change the `feed` object until the last step.

network_wrangler/transit/projects/add_route.py
```diff
@@ -89,71 +89,66 @@ def _add_route_to_feed(
...
-    if "shape_id" not in add_trips_df.columns:
-        add_trips_df["shape_id"] = None
-
-    if "trip_id" not in add_trips_df.columns:
-        add_trips_df["trip_id"] = None
-    add_trips_df["shape_id"] = fill_str_ids(add_trips_df["shape_id"], shapes_df["shape_id"])
-    add_trips_df["trip_id"] = add_trips_df["trip_id"].fillna(
-        add_trips_df["shape_id"].apply(lambda x: f"tr_shp{x}")
-    )
+    add_trips_df = pd.DataFrame([{
```
e-lo, line 146: @yueshuaing This is prescriptive such that only the specific properties you added will get added. the code I had on the left is generalized such that any value you added could be transferred which is consistent with GTFS model and our models as well.  Is there something that didn't work in the other method?

network_wrangler/transit/projects/add_route.py
```diff
@@ -89,71 +89,66 @@ def _add_route_to_feed(
...
-                [feed.stop_times, add_stop_times_df], ignore_index=True, sort=False
-            )
-
-    feed.stops = stops_df
-    feed.routes = routes_df
-    feed.shapes = shapes_df
-    feed.trips = trips_df
-    feed.stop_times = stop_times_df
-    feed.frequencies = frequencies_df
+            if route['agency_raw_name']:
```
e-lo, line 102: this requires that 'agency_raw_name' is used - which we shouldn't require

network_wrangler/transit/projects/add_route.py
```diff
@@ -284,7 +281,7 @@ def _create_new_frequencies(
         trip_id: Trip ID for the frequencies.
     """
     add_freqs_df = pd.DataFrame(
-        [_parse_headway_record(headway) for headway in headways],
+        [_parse_headway_record(headway)],
```
e-lo, line 281: this is just for a single headway record.  Don't you need as many records as there are time ranges?

e-lo (commented):
@yueshuaing I'm not sure about some of these changes - particuarly the ones that refer to agency_raw_name and lock the user into a specific set of properties. the code I wrote was agnostic to "extra" properties and would add them if provided.

network_wrangler/transit/projects/add_route.py
yueshuaing: Cube/Metcouncil wrangler uses `agency_raw_name` to do some data queries. @i-am-sijia Do we need to keep it?

network_wrangler/transit/projects/add_route.py
yueshuaing: I was just following the previous code, changing routes_df. Because I found the previous logic of updating trips_df, stops_df, etc., is not correct (as in each loop they're overwritten to reflect the latest changes), so I changed them to `feed`.

network_wrangler/transit/projects/add_route.py
yueshuaing: The previous code didn't add all trips as trips are added before frequencies. For example, we need to add four trips: direction 0 AM and MD, direction 1 AM and MD. The previous code only adds two trips of the two direction to the trips table. I changed the sequence, for each trip, check the headway (which also has the time of day info), then add new trip for each time period. I was trying to use the previous function, but since I changed the sequence of adding trips, some of the info like the shape_id, trip_id, has already been defined in the previous steps, so I simplified this function `_create_new_trips`

network_wrangler/transit/projects/add_route.py
yueshuaing: I added a loop [here](https://github.com/wsp-sag/network_wrangler/blob/selection-refactor/network_wrangler/transit/projects/add_route.py#L106) to add trip for each time period. so I changed the headways to headway

network_wrangler/scenario.py
```diff
@@ -604,224 +593,288 @@ def apply_projects(self, project_list: Collection[str]):
...
+def extract_base_scenario_metadata(base_scenario: dict) -> dict:
+    """Extract metadata from base scenario rather than keeping all of big files.
+
+    Useful for summarizing a scenario.
+    """
+    _skip_copy = ["road_net", "transit_net", "config"]
+    out_dict = {k: v for k, v in base_scenario.items() if k not in _skip_copy}
+    if isinstance(base_scenario.get("road_net"), RoadwayNetwork):
+        nodes_file = Path(base_scenario["road_net"].nodes_df.attrs["source_file"])
+        out_dict["roadway"] = {"dir": str(nodes_file.parent), "suffix": str(nodes_file.suffix)}
```
yueshuaing, line 815: Can't get the `nodes_file` path from nodes_df. The nodes_df.attrs is updated with the default settings [here](https://github.com/wsp-sag/network_wrangler/blob/selection-refactor/network_wrangler/roadway/links/io.py#L49-L50). Can we use links_df instead?

yueshuaing:
Do we have an example of config file that can be used to modify the ML_LINK_ID_METHOD?
I tried adding the two lines below in a yml file and specified the config parameter of create_scenario(). It didn't work.
```python
IDS:
    ML_LINK_ID_METHOD: "scalar"
    ML_LINK_ID_SCALAR: 2000000
```

network_wrangler/scenario.py
e-lo: There was a bug in the `validate_to_df()` that forgot to copy the `attr` properties at validation (from all the merging) that I fixed - can you confirm that it works now? 

WRT resolution, we shouldn't just change from nodes-->links. Need to fix if the `attr` is dropped.

e-lo:
> Do we have an example of config file that can be used to modify the ML_LINK_ID_METHOD?

Can you write a test? You should be able to just add a dict and follow directions here: https://wsp-sag.github.io/network_wrangler/selection-refactor/how_to/#change-wrangler-configuration

But if that isn't working then there is a bug we need to track down.

network_wrangler/transit/projects/add_route.py
e-lo: Ah - I think I remember now that this is function of how we decided to add new routes.  Can you add that to the documentation so it is clear?

network_wrangler/transit/projects/add_route.py
e-lo: Technically in GTFS there doesn't need to be a single trip/frequency.  Just a single trip per shape.  Do/should we require a single trip per frequency in our data model? If so, we need to enforce that in the data model.  Can you please create an issue if so @yueshuaing ?

network_wrangler/transit/projects/add_route.py
e-lo: you can keep it, but just either use generalized methods to perpetuate it into new routes or add it to cube/metcouncil logic

PR #329, opened 2023-06-14
Document network wrangler network format.

### LINKED ISSUE #278: [FEATURE] Documentation of roadway network schema and validation
### User Story
*As a modeler who builds networks, I would like to understand and (I'm guessing) take advantage of project card validation.*

I noticed our [script](https://github.com/BayAreaMetro/travel-model-two-networks/blob/develop/notebooks/02-all-projects.ipynb) which applies project cards is passing `validate=False`.  When I passed `validate=True`, I got errors because network_wrangler is trying to use yaml to load the non-yaml project cards (e.g. the `.wrangler` cards in this [folder](https://github.com/BayAreaMetro/travel-model-two-networks/tree/develop/project_cards).  I added a quick patch to skip those and I'm now getting this error:

> 2022-01-19 11:13:58, ERROR: Project Card Schema Loc:c:\users\lzorn\documents\github\network_wrangler\network_wrangler\schemas\project_card.json

So I was poking around the documentation/code/issues to understand what this schema is supposed to look like since it seems like we ought to create one but couldn't readily find much...
 
### Priority 

Medium ?

### Level of Effort

For a bit of documentation, probably not huge?

### Project

MTC 😄 

### Who should be involved?
Implementer: 
Commenters: 
Users: 
Reviewers: 

### Risk

#### Tests

e-lo:
- add documentation about network format
- lint

Fixes #278


e-lo (commented):
Just docs. Will merge.

Issue #330, opened 2023-11-03
[FEATURE] Update transit routing when roadway connection has changed


i-am-sijia:
### User Story
When users edit roadway network connections, e.g., delete a roadway link and replace it with new links, often time they forget to update the corresponding transit network for transit services run on top of the roadway links affected. And the commercial software platform that the users are making the network edits in do not automatically alert or reroute the transit due to the roadway changes. This causes the transit routes to be broken.

This is happening a lot in future year RTP network edits.

### Resolution Ideas
Add functionality in wrangler to check transit connectivity after applying highway projects, and reroute transit shapes if broken.

### Project
*Is there a funder or project associated with this feature?*

### Who should be involved?
Implementer: 
Commenters: 
Users: 
Reviewers: 

### Risk
*Will this potentially break anything?*

#### Tests
*What are relevant tests or what tests need to be created in order to determine that this issue is complete?*

PR #331, opened 2023-11-03
Fix broken transit shapes due to roadway changes

i-am-sijia:
Fix to issue #330 

The idea is to call the new method scenario.update_transit_net_with_new_road_net() after applying roadway project cards.


i-am-sijia:
@yueshuaing please test the changes.

PR #337, opened 2024-03-29
Apply transit line addition and deletion project cards

yueshuaing:
1. Add infrastructure to add New Transit Routes via project card (see schema here: https://github.com/network-wrangler/projectcard/issues/13). New records need to be added to the routes, trips, stops, stop_times, shapes, frequencies tables in the Standard Transit.
2. Delete transit lines via project card. Records will be removed from trips, stops, stop_times, frequencies tables in the Standard Transit.

### DIFF
--- a/network_wrangler/transitnetwork.py
+++ b/network_wrangler/transitnetwork.py
@@ -613,7 +613,7 @@ class TransitNetwork(object):
                     managed_lane_nodes,
                 )
             elif project_dictionary["category"].lower() == "add new route":
-                self.add_new_transit_feature_change(
+                self.add_new_transit_feature(
                     project_dictionary["routes"]
                 )
             elif project_dictionary["category"].lower() == "add transit":
@@ -1245,7 +1245,7 @@ class TransitNetwork(object):
             in_place=in_place,
         )
 
-    def add_new_transit_feature_change(
+    def add_new_transit_feature(
         self, routes: dict, in_place: bool = True
     ) -> Union(None, TransitNetwork):
         """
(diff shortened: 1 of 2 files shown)


network_wrangler/transitnetwork.py
```diff
@@ -1241,6 +1245,257 @@ def apply_transit_managed_lane(
...
+                    # used to add new stops if they are not in the stops.txt
+                    new_stop_id_list = []
+                    model_node_id_list = []
+                    stop_lat_list = []
+                    stop_lon_list = []
+
+                    for s in stop_model_node_id_list:
+                        s = int(float(s))
+                        if s in stop_id_xref_dict.keys():
+                            existing_agency_raw_name = stops_final_df[stops_final_df['model_node_id'].astype(float).astype(int)==s]['agency_raw_name'].iloc[0]
```
i-am-sijia, line 1426: Can existing `agency_raw_name` be more than one unique value? E.g., Metro and MVTA buses can use the same stop node?

i-am-sijia (changes requested):
@yueshuaing, Thank you. Overall the code change looks fine, please address my comments and we can merge it for now to move on with Met Council.

If we had more time, here are a few things we should do for this PR:

1. **Unit Testing.** When adding a new feature, we should add corresponding tests. See [here ](https://github.com/wsp-sag/network_wrangler/blob/bd90d435b6de83d070c0110ea199310d7410d63c/tests/test_transit_with_roadnet.py#L39-L111)for example. Ideally, we should have an example New Transit Route project card in the example folder and a test in the tests folder.

2. **Commenting and Docstring**. I want us to do better in adding in-line comments, especially for complicated and important features like this. Docstring should also be completed for documentation purpose. I saw you started a Docstring for `add_new_transit_feature_change`, which is great, but it's not complete. @yueshuaing Perhaps this is a low hanging fruit for you to address now quickly.

3. **Formatting**. I'd like to be more strict about code formatting, like [black](https://github.com/psf/black). We will force that with Network Wrangler v1.0.

i-am-sijia (approved):

PR #365, opened 2024-08-26
Update gtfs data model "stop_id" is a valid "model_node_id"

e-lo:
Instead of appending `model_node_id` and having a many to many relationship which violates GTFS, make `stop_id` a valid `model_node_id` and move original gtfs `stop_id` as aggregated string in `gtfs_stop_id`

Implementing this includes:

- Updating example data in `examples/...`
- Updating data models in `models/gtfs...`
- Updating functions and various merge operations that use nodes and stops
- Updating tests
- Adding helper utils in `models/gtfs/converters.py` (and tests for them) to migrate from original gtfs from ranch to wrangler gtfs.


e-lo:
@yueshuaing can you please make sure this works as expected for you and then we can merge in.

yueshuaing:
> @yueshuaing can you please make sure this works as expected for you and then we can merge in.

confirming that it functions as expected

yueshuaing:
confirming that it works as expected

PR #377, opened 2024-09-09
Fix: allow timespans over midnight

### LINKED ISSUE #370: Timespan is not in increasing order
## Describe the bug
This issue occurs when applying a transit_routing_change project card to change the route for the NT period which spans from19:00 to 3:00. The function [str_to_time_list](https://github.com/wsp-sag/network_wrangler/blob/selection-refactor/network_wrangler/utils/time.py#L167) convert this time span to [datetime.datetime(2024, 9, 5, 19, 0), datetime.datetime(2024, 9, 5, 3, 0)], triggering this error message
```python 
ValueError: Timespan is not in increasing order.
```

e-lo:
Fixes #370 such that if timespans are not in increasing order, assume that they span midnight.

Note parsing of times > 24 hrs still doesn't work :-/

### DIFF
--- a/network_wrangler/models/_base/series.py
+++ b/network_wrangler/models/_base/series.py
@@ -1,9 +1,11 @@
 import pandera as pa
 
-# Define a general schema for any Series with time strings in HH:MM or HH:MM:SS format
+"""
+Time strings in HH:MM or HH:MM:SS format up to 48 hours.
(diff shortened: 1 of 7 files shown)


e-lo:
@yueshuaing confirm this addresses your issue

PR #378, opened 2024-09-12
Instrument + address performance

### LINKED ISSUE #363: Runtime increase when applying project cards
We observed a significant increase in runtime when applying project cards. Previously, it took about 30 minutes to apply all project cards in [BaseCorrections](https://github.com/Metropolitan-Council/project_card_registry/tree/Test1/projects/BaseCorrections) using the old NW, but now it takes 120 minutes in new NW. We found two potential reasons after a brief review of the code:

1. Performing validation checks each time a new change is applied.
2. The get_selection() method is taking longer to select links.

e-lo:
- Reduce multiple validations of same data
- Add instrumentation of performance 

Fixes #363


e-lo:
@i-am-sijia I think you can approve this and then we can start instrumenting our PRs in general.

yueshuaing:
@i-am-sijia and @e-lo, I tested this branch with a small set of project cards, the runtime looks shorter. I'll merge this PR so that I can continue testing applying all project cards with other issues fixed. Let me know if you have comments.

Issue #406, opened 2025-07-19
🧹 Use `default_factory` for mutable default


i-am-sijia:
`import network_wrangler` in newer python versions (I believe >3.10) gives error:

```python
File ~\AppData\Local\Programs\Python\Python313\Lib\site-packages\network_wrangler\configs\wrangler.py:210
    193     """CPU Configuration -  Will not change any outcomes.
    194 
    195     Attributes:
    196         EST_PD_READ_SPEED: Read sec / MB - WILL DEPEND ON SPECIFIC COMPUTER
    197     """
    199     EST_PD_READ_SPEED: dict[str, float] = Field(
    200         default_factory=lambda: {
    201             "csv": 0.03,
   (...)    206         }
    207     )
--> 210 @dataclass
    211 class WranglerConfig(ConfigItem):
    212     """Configuration for Network Wrangler.
    213 
    214     Attributes:
   (...)    218         EDITS: Parameters governing how edits are handled.
    219     """
    221     IDS: IdGenerationConfig = IdGenerationConfig()

ValueError: mutable default <class 'network_wrangler.configs.wrangler.IdGenerationConfig'> for field IDS is not allowed: use default_factory
```

### Ideas for resolution
Use `IDS: IdGenerationConfig = Field(default_factory=IdGenerationConfig())`?

e-lo:
Sounds like it would fix it - I will merge the PR once the tests have been udpated - want to take that on @diego-galdino ?

PR #407, opened 2025-07-19
Use `default_factory` for mutable default

i-am-sijia:
Addresses #406


e-lo:
#429 should subsume this

PR #429, opened 2026-01-06
Hotfix: Pandera/Pandas Compatibility and Python 3.10+ Compatibility

e-lo:
## Problem

The package failed to import with `TypeError: dtype '<class 'list'>' not understood `due to a compatibility issue between Pydantic's validate_call decorator and Pandera's DataFrame type hints when schemas contain list types.

Additionally, there were:
- A dataclass mutable default argument issue in `WranglerConfig`
- Python 3.10+ compatibility issues with Pydantic v2 field validators
- A bug in `Feed.__init__` where dict iteration was missing .items()

## Solution

This hotfix addresses all compatibility and bug issues:
- Fixed Pandera/Pydantic compatibility (`scopes.py`, `network.py`)
- Changed `DataFrame[Model]` type annotations to `pd.DataFrame` in function signatures and class fields
- Added explicit validation using `validate_df_to_model()` to maintain the same validation behavior
- Added `validate_return=False` to `@validate_call` decorators where needed
- Fixed dataclass mutable default (`wrangler.py`)
- Changed WranglerConfig fields to use `Field(default_factory=...)` instead of direct instantiation
- Removed unnecessary helper functions by using class constructors directly
- Fixed Python 3.10+ compatibility (`network.py`, `feed.py`)
- Fixed @field_validator decorator order (classmethod before `field_validator`) for Pydantic v2
- Fixed dict iteration bug in `Feed.__init__`

## Testing

✅ Package imports successfully without errors
✅ 242 tests passing (excluding benchmarks/docs which require additional setup)
✅ All core functionality tests pass
✅ Validation still occurs (moved from implicit to explicit)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -39,8 +39,6 @@ plugins:
       handlers:
         python:
           paths: [.]
-          selection:
-            new_path_syntax: true
           rendering:
             show_root_heading: true
             show_source: true


e-lo:
@lmz and @i-am-sijia - it is taking a lot of effort to get the backwards compatibility for documentation to work with Python 3.9. That said, we don't actually need the documentation to work with Python 3.9 (previously it was being run just with 3.10). All the tests pass with 3.9 thru 3.13, so I think PR is safe to merge to `main` and will fix immediate later python compatability issues. I will submitseparate PRs for deprecating python 3.9 support #430 and separating out our documentation builds from other testing #432 .

lmz:
This looks fine to me; I think we should just merge it in.

lmz (approved):

PR #478, opened 2026-06-19
Add functionality to create centroids and centroid connector to model network

lmz:
This PR adds end-to-end support for creating centroids and centroid connectors in a model network, including the core APIs, geometry/link generation logic, and tests/documentation updates needed to use and validate the workflow.

### What this PR does
This PR introduces functionality to build model-zone access structures by generating:

- Centroid nodes (zone representation in the network)
- Centroid connector links between centroids and nearby network nodes
- The feature is designed to support common modeling workflows where zones/TAZs need to be integrated into the routable network with generated access/egress links.

### Key changes

- Added APIs/utilities to create centroid nodes from input zone geometry/data.
- Added logic to create centroid connector links from each centroid to eligible nearby network nodes.
- Added validation and guardrails around connector creation inputs and outputs.
- Added/updated tests covering centroid and connector creation behavior.
- Updated docs/examples to demonstrate expected usage and outputs.

### Why this change is needed
Many demand and assignment workflows require zone-level connectors to represent how trips enter/exit the physical network. This PR provides a native, repeatable way to generate those structures directly in network_wrangler, reducing manual preprocessing and ensuring consistent connector generation behavior.

### Implementation notes
Centroid generation and connector generation are implemented as explicit network-building operations so they can be run as part of a reproducible wrangling pipeline.
Connector creation is based on proximity/eligibility rules to attach centroids to nearby nodes.
The new functionality is intended to be configurable for model-specific connector policies.

### Testing

This PR includes tests for:
- Correct centroid creation from valid inputs
- Connector generation for expected nearest/eligible nodes
- Error/edge handling for invalid or missing inputs
- Basic integrity checks on produced nodes/links
- Backward compatibility

Existing workflows are unaffected unless the new centroid/connector functionality is explicitly invoked. No breaking changes are intended for users not using these new APIs.

### Limitations / follow-ups

Connector selection policies may still need tuning for specific model conventions (e.g., max connectors per zone, facility/type filtering, directional behavior).
Future enhancements can expand configurability and performance for very large zone sets.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1215523318902339


lmz:
cc @i-am-sijia

docs/design.md
```diff
@@ -30,6 +30,127 @@ NetworkWrangler deals with four primary atomic parts:
 
 **4.[`ProjectCard`](https://network-wrangler.github.io/projectcard/main/api/#projectcard.projectcard.ProjectCard)** objects store information (including  metadata) about changes to the network.  Network Wrangler uses the [`projectcard`](https://network-wrangler.github.io/projectcard) package to read project cards from .yaml-like files and validate them.
 
+## Creating Networks
```
e-lo: most of this belongs in: https://github.com/network-wrangler/network_wrangler/blob/main/docs/how_to.md

network_wrangler/roadway/links/create.py
```diff
@@ -109,7 +122,38 @@ def data_to_links_df(
     links_df.attrs.update(RoadLinksAttrs)
     links_df = set_df_index_to_pk(links_df)
     links_df.gdf_name = links_df.attrs["name"]
+
+    # Ensure ML_geometry is a proper GeoSeries before validation.
+    # pandera >= 0.26 rejects a plain object column of None as not geometry dtype,
+    # even with nullable=True and coerce=True. Pre-casting to GeoSeries avoids this.
+    # Values loaded from GeoJSON files may be dicts or JSON strings; convert to Shapely first.
+    if "ML_geometry" not in links_df.columns:
```
e-lo: 1. Do we want ML_geometry to be mandatory?
2. This coercing should be its own function (e.g. `_fill_missing_link_geometries_from_nodes()`)
3. All the `apply()` is very SLOW...especially when we are using it just for counting. 

Suggest 
(a) just doing the counting to make sure things work for a **test** on a tiny network (not here).  
(b) i think you can do a select just on ones that don't work (if any) and then apply? If its null should be skipping entirely.

network_wrangler/roadway/nodes/create.py
```diff
@@ -140,6 +140,27 @@ def _create_nodes_from_link(
     )
     nodes_df["X"] = nodes_df.geometry.x
     nodes_df["Y"] = nodes_df.geometry.y
+
+    # Handle duplicate model_node_id values by averaging coordinates
```
e-lo: 1. this should be an option rather than default - in many (most?) cases this indicates an *error* which shouldn't be obfuscated.
2. should encapsulate in a function "merge nodes"
3. should probably error/warn if distances between are really big (i.e. almost definitely indicates an error)

network_wrangler/roadway/links/connectors.py
```diff
@@ -0,0 +1,473 @@
+"""Functions to create centroid connectors."""
+
```
e-lo, line 18: Add usage

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
+"""Functions to create centroid connectors."""
```
e-lo: Should this be in a sub-module similar to /nodes /links?

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    OKAY = 3
+    DO_NOT_USE = 100
+
+
+def calculate_angle_from_centroid(
+    gdf: gpd.GeoDataFrame,
+    centroid_col: str = "geometry_centroid",
+    angle_col: str = "angle_from_north",
+) -> gpd.GeoDataFrame:
+    """Calculate the angle from centroid to point, measured clockwise from true north.
```
e-lo: aka "bearing"

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        centroid_col: Name of column containing the centroid geometry (default: 'geometry_centroid')
+        angle_col: Name of the new column to create with angle values (default: 'angle_from_north')
+
+    Returns:
+        GeoDataFrame with new angle column added
+
+    Note:
+        - Angle is measured clockwise from north (0° = north, 90° = east, 180° = south, 270° = west)
+        - Returns values in range [0, 360)
+        - Assumes geometries are in a projected coordinate system or handles lat/lon appropriately
```
e-lo: this sounds ambiguous. if it assumes its projected then it shouldn't work for lat/long?

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    """
+
+    def calculate_bearing(centroid, point):
```
e-lo: make this generic between two points and add to generic geo utils.

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        angle_deg = math.degrees(angle_rad)
+
+        # Convert from east-based to north-based (geographic convention)
+        # and from counter-clockwise to clockwise
+        bearing = (90 - angle_deg) % 360
+
+        return bearing
+
+    # Calculate angles for all rows
+    gdf[angle_col] = gdf.apply(
```
e-lo: apply is slow! use vectorized arctan w/numpy: 

```python
dx = gdf.geometry.x.to_numpy() - gdf[centroid_col].x.to_numpy()
dy = gdf.geometry.y.to_numpy() - gdf[centroid_col].y.to_numpy()
gdf[angle_col] = (90 - np.degrees(np.arctan2(dy, dx))) % 360
```

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        lambda row: calculate_bearing(row[centroid_col], row["geometry"]), axis=1
+    )
+
+    return gdf
+
+
+def add_centroid_nodes(
```
e-lo: this seems like a separate (first) step which shoudl be in `/nodes` submodule named `centroids.py` (and the other stuff should pprobably be `connectors.py`  under `/links`)

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    zones_gdf: gpd.GeoDataFrame,
+    zone_id: str,
+    default_node_attribute_dict: dict[str, any] | None = None,
+):
+    """Adds the given centroid nodes to the roadway network.
+
+    Args:
+        road_net: the RoadwayNetwork to update by adding centroids
+        zones_gdf: zones definition which must have two geometry columns:
+            'geometry', which the geometry boundary, and 'centroid_geometry',
```
e-lo: need to define this better. is there a `zone_id` column? what is its type? should define a Zones data model somehwere!

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            which should contain the centroid point location
+        zone_id: the zone id field in zones_gdf; this will be used as the
```
e-lo: "used for the model_node_id"...+"for the centroid that represents that zone"

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            model_node_id
+        default_node_attribute_dict: node attributes to set for the new centroid nodes.
```
e-lo: what is an example of how this would be used?

e-lo: Also are there other attributes that would be copied from zone_gdf  columns?

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            Defaults to None.
+    """
+    centroid_nodes_gdf = (
+        zones_gdf[[zone_id, "geometry_centroid"]]
+        .rename(columns={"geometry_centroid": "geometry", zone_id: "model_node_id"})
+        .set_geometry("geometry", crs=LAT_LON_CRS)
+    )
+    centroid_nodes_gdf["X"] = centroid_nodes_gdf["geometry"].x
+    centroid_nodes_gdf["Y"] = centroid_nodes_gdf["geometry"].y
+    centroid_nodes_gdf["osm_node_id"] = f"{zone_id}:" + centroid_nodes_gdf["model_node_id"].astype(
```
e-lo: this isn't an osm_node_id though...why create a fake one? should use a different field.

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        str
+    )
+
+    # set default node attributes
+    if default_node_attribute_dict is None:
```
e-lo: next 4 lines can just be: 

```python
... = centroid_nodes_gdf.assign(**(default_node_attribute_dict or {}))
```

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        default_node_attribute_dict = {}
+    for colname, default_value in default_node_attribute_dict.items():
+        centroid_nodes_gdf[colname] = default_value
+
+    # assume the model_node_id
+    len_road_net_nodes = len(road_net.nodes_df)
+    WranglerLogger.debug(f"centroid_nodes_gdf:\n{centroid_nodes_gdf}")
+    road_net.add_nodes(centroid_nodes_gdf)
+    WranglerLogger.info(
+        f"Added node centroids for {zone_id}: "
```
e-lo: isn't this just the zone-Id field name?

network_wrangler/roadway/links/connectors.py
```diff
@@ -0,0 +1,473 @@
...
+    len_road_net_nodes = len(road_net.nodes_df)
+    WranglerLogger.debug(f"centroid_nodes_gdf:\n{centroid_nodes_gdf}")
+    road_net.add_nodes(centroid_nodes_gdf)
+    WranglerLogger.info(
+        f"Added node centroids for {zone_id}: "
+        f"increased size of nodes_df from {len_road_net_nodes:,} to {len(road_net.nodes_df):,}"
+    )
+
+
+def add_centroid_connectors(  # noqa: PLR0912, PLR0915
```
e-lo: If you have a really long doc string with paragraph-long steps, you should probably be breaking up the function into that number of steps :-)

Something like: 

1. evaluate_nodes (finds good nodes, filters out unusable ones)
2. evaluate_connectors (evaluate how well each node works with each zone)
3. select_connectors (from candidates, what is best)
4. create_connectors
5. add connectors to network

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    Returns:
+        A copy of zones_gdf with an additional column, `num_connectors`. The road_net is
+            updated in place with new centroid connector links, and the nodes table has an
+            additional column: `{zone_id}_num_connectors`.
+
+    """
+    WranglerLogger.info(f"Adding centroid connectors for zone:{zone_id} and mode:{mode}")
+    WranglerLogger.debug(f"zones_gdf:\n{zones_gdf}")
+
+    G = road_net.get_modal_graph(mode)
```
e-lo: Super expensive operation when you are just getting degrees - not doing anything like a path.

Would be much faster to just do a merge on A and group-by.count something like:

```python
scores = modal_links.groupby("A")[f"{mode}_centroid_fit"].agg(
    **{f"{mode}_graph_degrees": "size", f"{mode}_centroid_fit": "max"}
)
```

And then filter to usable ones:

```python
fit_scores = scores[scores[fit_col] != FitForCentroidConnection.DO_NOT_USE]
```

And then attach that info to nodes:

```python
candidate_nodes = nodes_df.merge(
        scores,
        left_on="model_node_id",
        right_index=True,
        how="inner",
        validate="one_to_one",
    )
```

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    # filter to usable nodes
+    mode_node_df = mode_node_df.loc[
+        mode_node_df[f"{mode}_centroid_fit"] != FitForCentroidConnection.DO_NOT_USE
+    ]
+
+    # get node information for these nodes
+    mode_node_df = gpd.GeoDataFrame(
+        pd.merge(
+            left=mode_node_df,
+            right=road_net.nodes_df[["model_node_id", "geometry", "osm_node_id", "street_count"]],
```
e-lo: street_count?? why select at all and not import all values?

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    WranglerLogger.debug(f"Created road_net modal_graph for {mode}:")
+
+    node_dict = {}  # node -> *outgoing* degree to all nodes, centroid_fit
+    for node in G.nodes():
+        # calculate each nodes fitness for centroid based on max of links
+        edge_fit_values = [data[f"{mode}_centroid_fit"] for u, v, data in G.edges(node, data=True)]
+        # WranglerLogger.debug(f"node: {node} edge_fit_values:{edge_fit_values}")
+        node_dict[node] = [G.out_degree(node), max(edge_fit_values) if edge_fit_values else None]
+
+    mode_node_df = pd.DataFrame.from_dict(
```
e-lo: this df name is confusing

network_wrangler/utils/geo.py
```diff
@@ -597,3 +625,108 @@ def update_point_geometry(
         fail_if_missing=False,
     )
     return updated_df
+
+
+def get_link_bearing_degrees(geometry) -> float:
```
e-lo: Guessing this is something that you will do for a whole geoseries - so best to vectorized operation (see other comment with algorithm) using numpy

network_wrangler/roadway/centroids.py
e-lo: see above comment for more efficient way to do this...

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        geometry="geometry",
+        crs=LAT_LON_CRS,
+    )
+
+    mode_node_df.fillna({f"{mode}_graph_degrees": 0}, inplace=True)
+    mode_node_df.fillna(
+        {f"{mode}_centroid_fit": FitForCentroidConnection.DO_NOT_USE}, inplace=True
+    )
+    mode_node_df[f"{mode}_graph_degrees"] = mode_node_df[f"{mode}_graph_degrees"].astype(int)
+    mode_node_df[f"{mode}_centroid_fit"] = mode_node_df[f"{mode}_centroid_fit"].astype(int)
```
e-lo: these are always int64s

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    # be connected to multiple centroid connectors
+    WranglerLogger.debug(
+        f"After spatial join, non-unique model_node_ids:\n"
+        f"{mode_node_df.loc[mode_node_df['model_node_id'].duplicated(keep=False)]}"
+    )
+
+    # calculate distance from centroid
+    gs = gpd.GeoSeries(mode_node_df["geometry_centroid"], crs=LAT_LON_CRS)  # source CRS
+    mode_node_df["geometry_centroid"] = gs.to_crs(local_crs).values  # target CRS
+    mode_node_df["distance_from_centroid"] = mode_node_df.apply(
```
e-lo: ak - another apply 👎 (unless you need a long coffee break).
The `.distance` method can be vectorized.

```python
cent_geom = gpd.GeoSeries(mode_node_df["geometry_centroid"], crs=LAT_LON_CRS).to_crs(local_crs)
mode_node_df["distance_from_centroid"] = mode_node_df.geometry.distance(cent_geom, align=False)
```

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    # add angle fron centroid
+    mode_node_df = calculate_angle_from_centroid(
+        mode_node_df, "geometry_centroid", "centroid_angle"
+    )
+    WranglerLogger.debug(
+        f"After adding angle from centroid, mode_node_df type={type(mode_node_df)}:\n{mode_node_df}"
+    )
+
+    # Filter to nodes within the given zones
+    mode_node_df = mode_node_df.loc[mode_node_df[zone_id].isna() == False]
```
e-lo: .notna()

network_wrangler/roadway/links/connectors.py
```diff
@@ -0,0 +1,473 @@
...
+        1. **Node Evaluation**: For each node in the modal graph, calculates:
+            - Outgoing degree (number of outbound links)
+            - Fitness for centroid connection (worst fitness of connected links)
+
+        2. **Spatial Filtering**: Identifies nodes within each zone boundary and filters out:
+            - Nodes with `{mode}_centroid_fit` = DO_NOT_USE
+            - Nodes outside zone boundaries
+            - Nodes with outgoing degree > max_mode_graph_degrees
+
+        3. **Connector Selection** (per zone):
```
e-lo: this is REALLY inefficient! O(zones × connectors × candidates).  


Can fix to O(zones x connectors) by vectorizing some apply() statements (suggestions below)

e-lo: OR

Could consider using an algorithm which would pick something per sector (admittedly Claude suggested this approach as quicker):

```python
n = num_centroid_connectors
w = 360.0 / n
df["sector"] = ((df["centroid_angle"] % 360) // w).astype(int)
df.sort_values([zone_id, fit_col, "distance_from_centroid"], inplace=True)
df["sector_rank"] = df.groupby([zone_id, "sector"]).cumcount()
df.sort_values([zone_id, "sector_rank", fit_col, "distance_from_centroid"], inplace=True)
selected = df.groupby(zone_id).head(n).copy()
selected["connector_num"] = selected.groupby(zone_id).cumcount() + 1
```

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            WranglerLogger.warning(f"No centroid connectors for {zone_id} {zone_num}")
+            continue
+
+        # 1: choose the connector with the lowest f"{mode}_centroid_fit" and lowest distance_from_centroid
+        # Since data is already sorted by fit and distance, first row is the best
+        first_idx = zone_data.index[0]
+        mode_node_df.loc[first_idx, "connector_num"] = 1
+        zone_data.loc[first_idx, "connector_num"] = 1
+
+        if zone_num == 650:  # noqa: PLR2004
```
e-lo: debug leftover? shouldn't be in here as magic # really, right?

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            # Among candidates with the best fit, choose the one with maximum angular separation
+            if len(best_fit_candidates) > 0:
+                selected_idx = best_fit_candidates["min_angle_sep"].idxmax()
+                mode_node_df.loc[selected_idx, "connector_num"] = connector_num
+                # Update zone_data to reflect this selection for next iteration
+                zone_data.loc[selected_idx, "connector_num"] = connector_num
+            else:
+                break  # No more candidates
+
+        if zone_num == 650:  # noqa: PLR2004
```
e-lo: ??

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+            candidates["min_angle_sep"] = candidates["centroid_angle"].apply(
+                lambda angle, _sa=selected_angles: calculate_min_angle_separation(angle, _sa)
+            )
+
+            # Group by centroid_fit level and find the one with max angular separation within each level
+            # Since data is already sorted by fit, we can process in order
+            best_fit = candidates[f"{mode}_centroid_fit"].min()
+            best_fit_candidates = candidates[candidates[f"{mode}_centroid_fit"] == best_fit]
+
+            if zone_num == 650:  # noqa: PLR2004
```
e-lo: ???

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+
+            if not candidate_mask.any():
+                break  # No more candidates for this zone
+
+            selected_angles = zone_data.loc[selected_mask, "centroid_angle"].values
+            candidates = zone_data[candidate_mask].copy()
+
+            # Calculate minimum angular separation for each candidate
+            candidates["min_angle_sep"] = candidates["centroid_angle"].apply(
+                lambda angle, _sa=selected_angles: calculate_min_angle_separation(angle, _sa)
```
e-lo: This is an expensive apply which re-scans the whole selected set each time.

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+
+            if zone_num == 650:  # noqa: PLR2004
+                WranglerLogger.debug(
+                    f"B {zone_id} {zone_num} {connector_num=}:\n{best_fit_candidates}"
+                )
+
+            # Among candidates with the best fit, choose the one with maximum angular separation
+            if len(best_fit_candidates) > 0:
+                selected_idx = best_fit_candidates["min_angle_sep"].idxmax()
+                mode_node_df.loc[selected_idx, "connector_num"] = connector_num
```
e-lo: `connector_num` is mutated in different and potentially conflicting places and might get out of sync.

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+    ]
+
+    # sort by drive_centroid_fit, centroid_angle
+    mode_node_df.sort_values(
+        by=[zone_id, f"{mode}_centroid_fit", "distance_from_centroid"], inplace=True
+    )
+    mode_node_df.reset_index(drop=True, inplace=True)
+    mode_node_df["connector_num"] = 0
+
+    def calculate_min_angle_separation(candidate_angle, selected_angles):
```
e-lo: This helper can be removed...see below

network_wrangler/roadway/centroids.py
```diff
@@ -0,0 +1,473 @@
...
+        angle_diffs = np.abs(candidate_angle - selected_angles)
+        # Handle wraparound
+        angle_diffs = np.minimum(angle_diffs, 360 - angle_diffs)
+        return np.min(angle_diffs)
+
+    WranglerLogger.debug(
+        f"Before choosing centroid connector nodes, mode_node_df:\n{mode_node_df}"
+    )
+
+    # Process each zone and select connectors
```
e-lo: I was pretty sure there was a way to do this without the extra applies and this is what Claude came up with which I think is ok.  Its basically just keeping a running tab of the min angle of already selected connectors for each zone and updating as they are selected. I only 68% understand this (but didn't burn a ton of time on it either) but it should be faster. 

```python
fit_col = f"{mode}_centroid_fit"
# rows are already sorted by [zone_id, fit_col, distance_from_centroid], so within a
# zone the first row is the best fit + closest.
for zone_num in mode_node_df[zone_id].unique():
    idx = mode_node_df.index[mode_node_df[zone_id] == zone_num]
    if len(idx) == 0:
        continue

    angles = mode_node_df.loc[idx, "centroid_angle"].to_numpy()
    fit = mode_node_df.loc[idx, fit_col].to_numpy()
    conn = np.zeros(len(idx), dtype=int)

    def circ_sep(a):  # wraparound angular distance of every angle from scalar a
        d = np.abs(angles - a)
        return np.minimum(d, 360 - d)

    conn[0] = 1                       # connector 1: best fit + closest
    min_sep = circ_sep(angles[0])     # running min separation to the selected set

    for k in range(2, num_centroid_connectors + 1):
        unsel = conn == 0
        if not unsel.any():
            break
        best_fit = fit[unsel].min()               # best remaining fit level
        eligible = unsel & (fit == best_fit)
        pick = np.argmax(np.where(eligible, min_sep, -np.inf))
        conn[pick] = k
        min_sep = np.minimum(min_sep, circ_sep(angles[pick]))   # incremental, O(n)

    mode_node_df.loc[idx, "connector_num"] = conn
```

network_wrangler/roadway/model_roadway.py
```diff
@@ -446,9 +537,15 @@ def _create_dummy_connector_links(
         raise ManagedLaneAccessEgressError(msg)
 
     # access link should go from A_GP to A_ML
+    # Use larger offsets (1M and 2M) to avoid collisions when GP link IDs differ by small amounts
```
e-lo: this should probably be a parameter and not hard coded...right? there are different limits in various softwares.

network_wrangler/roadway/network.py
```diff
@@ -1019,6 +1021,43 @@ def is_connected(self, mode: str) -> bool:
 
         return is_connected
 
+    def write(
```
e-lo, line 1024: if do this here - should do it to transit!?

network_wrangler/utils/geo.py
```diff
@@ -597,3 +625,108 @@ def update_point_geometry(
...
+    bearing_rad = get_bearing(lat1, lon1, lat2, lon2)
+
+    # Convert to degrees and normalize to 0-360
+    bearing_deg = math.degrees(bearing_rad)
+    bearing_deg = (bearing_deg + 360) % 360
+
+    return bearing_deg
+
+
+def bearing_to_cardinal_direction(bearing: float, cardinal_only: bool = False) -> Optional[str]:
```
e-lo: do vector operations applied to series...

```python
def bearings_to_cardinal_directions(bearings: pd.Series, cardinal_only: bool = False) -> pd.Series:
    """Vectorized bearing (deg) -> cardinal/intercardinal label. NaN bearings map to None."""

    _CARDINAL_LABELS = np.array(
    [d.value for d in (CardinalDirection.N, CardinalDirection.E, CardinalDirection.S, CardinalDirection.W)]
) #also do intercardinal if want that....

    labels, sector_deg = (
        (_CARDINAL_LABELS, 90.0) if cardinal_only else (_INTERCARDINAL_LABELS, 45.0)
    )
    # Shift by half a sector so North straddles 0°, then floor-divide into sector indices.
    sector = ((bearings + sector_deg / 2) % 360) // sector_deg

    directions = pd.Series(pd.NA, index=bearings.index, dtype="object")
    valid = sector.notna()
    directions[valid] = labels[sector[valid].astype(int)]
    return directions
```

e-lo (changes requested):
Some things to address :-)

lmz:
Thank you @e-lo !  I'll go through these ASAP, hopefully by end of week!

docs/design.md
lmz: Fixed in 05ec59e696663a4b9538cc7c7f871aee84ecea68

network_wrangler/roadway/centroids.py
lmz: Fixed in [6016b6b5d2df1124ff83ba8cebb6b2399bf715c8](https://github.com/BayAreaMetro/network_wrangler/commit/6016b6b5d2df1124ff83ba8cebb6b2399bf715c8)

network_wrangler/roadway/centroids.py
lmz: Fixed in 0b9357e076279eff8661dc965d0c6e10bf4c9bde

network_wrangler/roadway/centroids.py
lmz: 🫠 Sorry! 75a90cffa4652fa8977ed584fb19eb6cecb1e13d

network_wrangler/roadway/centroids.py
lmz: Doh! 75a90cffa4652fa8977ed584fb19eb6cecb1e13d

network_wrangler/roadway/centroids.py
lmz: Yes, sorry! https://github.com/network-wrangler/network_wrangler/commit/75a90cffa4652fa8977ed584fb19eb6cecb1e13d

network_wrangler/utils/geo.py
lmz: Good point! vectorized.

network_wrangler/roadway/links/create.py
lmz: Got it. I'm not convinced ML_geometry should be mandatory...
Moved the coercion to its own function, `_prepare_ml_geometry_for_validation()` and only coerce rows where it's actually needed.

network_wrangler/roadway/links/connectors.py
lmz: Added in 76954b541030f260f994d4bd507bec402c464e23

network_wrangler/roadway/centroids.py
lmz: Removed.

network_wrangler/roadway/centroids.py
lmz: Makes sense; updated!

network_wrangler/roadway/centroids.py
lmz: Fixed.

network_wrangler/roadway/centroids.py
lmz: Fixed; it's now `point_bearings_degrees()` in utils\geo.py

network_wrangler/utils/geo.py
lmz: Fixed.

network_wrangler/roadway/centroids.py
lmz: Ok :D I just liked using it because it's there. Removed in 2e2558e30ab0c781bdce1ede29f2da55d5221d1c

network_wrangler/roadway/centroids.py
lmz: Clarified in 2e2558e30ab0c781bdce1ede29f2da55d5221d1c

network_wrangler/roadway/model_roadway.py
lmz: Right.  Fixed in 4f867aaf957ef53972f53e07aaafd4822d1e5745

network_wrangler/roadway/nodes/create.py
lmz: Encapsulated into function and issue warning in f2cdd80ff00d7cd06e0af1a88a7c146e0221e939

network_wrangler/roadway/centroids.py
lmz: Defined initial version of ZonesTable in 7080321d11b547976333bb114cb6181e16147d39

network_wrangler/roadway/centroids.py
lmz: I am using it for default custom node attributes: https://github.com/BayAreaMetro/mtc_wrangler/blob/db19cc47da7b235aaf5ee003952fca97f203589b/models/mtc_roadway_schema.py#L230

network_wrangler/roadway/centroids.py
lmz: I tried to clarify the zone_id_col as an attribute in the ZonesTable so you could have a TAZ ZonesTable and a MAZ ZonesTable both.

network_wrangler/roadway/centroids.py
lmz: Good point! fixed.

network_wrangler/roadway/centroids.py
lmz: Good point! fixed.

e-lo:
I see you are hard at work on this - LMK when you are ready for another review :-)

network_wrangler/roadway/centroids.py
lmz: Renamed 😄

network_wrangler/roadway/centroids.py
lmz: Removed in f4aec9fb9a5d1aa883c509046a6a0bc2f576c628

network_wrangler/roadway/centroids.py
lmz: Fixed.

network_wrangler/roadway/centroids.py
lmz: Renamed to candidate_nodes_df

network_wrangler/roadway/links/connectors.py
lmz: Good point! Fixed in 72a1aa81700ed224c0e239645e631a2b20dc761b

network_wrangler/roadway/centroids.py
lmz: Fixed

network_wrangler/roadway/centroids.py
lmz: Fixed

network_wrangler/roadway/centroids.py
lmz: Reasonable. Fixed in 158d4b2f7f7be0380acabe9d2a20381979d082ae

network_wrangler/roadway/centroids.py
lmz: Split it up into centroids.py and connectors.py

network_wrangler/roadway/centroids.py
lmz: Good point .Removed the "or handles lat/lon appropriately" hedge and replaced it with an explicit ValueError guard: the function now raises immediately if the input gdf has a geographic (lat/lon) CRS.

network_wrangler/roadway/links/connectors.py
lmz: Broke it up in 4079fd5b7fc3e87a5d746bba40bbb9acb3a1d43d

lmz:
> I see you are hard at work on this - LMK when you are ready for another review :-)

I think I've addressed your comments and it's ready for another look!
Thank you!! (And I'm still pushing hard on contracting...)

network_wrangler/roadway/nodes/centroids.py
```diff
@@ -0,0 +1,116 @@
...
+    validates/coerces against ``ZonesTable``.
+
+    Metadata is preserved in ``attrs`` and can be augmented with ``metadata``.
+    The original zone-id source column name is stored as ``attrs['zone_id_col']``.
+    """
+    if zone_id_col not in zones_gdf.columns:
+        msg = f"zones_gdf is missing required zone id column: {zone_id_col}"
+        raise ValueError(msg)
+
+    normalized = zones_gdf.rename(columns={zone_id_col: "zone_id"}).copy()
```
e-lo: what is the purpose of creating this if you are then dumping it? its cleaner IMHO to deal with the normalized table then to deal with setting and having to keep track of table attributes for column names.

network_wrangler/roadway/nodes/centroids.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def prepare_zones_table(
+    zones_gdf: gpd.GeoDataFrame,
+    zone_id_col: str,
+    metadata: dict[str, object] | None = None,
+) -> DataFrame[ZonesTable]:
+    """Create a validated zones table for centroid workflows.
+
+    Renames the user-specified ``zone_id_col`` to ``zone_id`` and
```
e-lo: right now its not renaming it to zone_id (but....it should IMHO! why wouldn't it?)

network_wrangler/models/roadway/tables.py
```diff
@@ -400,6 +401,31 @@ class Config:
         _pk: ClassVar[TablePrimaryKeys] = ["shape_id"]
 
 
+class ZonesTable(DataFrameModel):
+    """Datamodel for zone polygons and centroid points used by centroid connectors.
+
+    Attributes:
+        zone_id (int): Zone identifier used as centroid ``model_node_id``.
```
e-lo: right now you aren't validating the normalized table so unless your zone_id *happens* to be zone_id then this will fail. THat siad, I'd prefer you fix this in the prepare zonal data file step and actually use the normalized data.

network_wrangler/roadway/nodes/create.py
```diff
@@ -140,13 +140,64 @@ def _create_nodes_from_link(
...
+    dupes_projected["y_m"] = dupes_projected.geometry.y
+
+    spread_by_id = (
+        dupes_projected.groupby("model_node_id")["geometry"]
+        .apply(lambda geom: geom.apply(lambda g: geom.distance(g).max()).max())
+        .astype(float)
+    )
+    wide_spread = spread_by_id[spread_by_id > DUPLICATE_NODE_WARN_SPREAD_METERS]
+    if not wide_spread.empty:
+        WranglerLogger.warning(
```
e-lo: do we want a warning or an error? if it is truly a WIDE SPREAD then i think this should fail

network_wrangler/roadway/nodes/create.py
```diff
@@ -140,13 +140,64 @@ def _create_nodes_from_link(
     )
     nodes_df["X"] = nodes_df.geometry.x
     nodes_df["Y"] = nodes_df.geometry.y
+
+    nodes_df = _merge_duplicate_nodes_by_average(nodes_df)
```
e-lo: this should be an opt-in.  In * most networks * this indicates an error.

network_wrangler/roadway/nodes/create.py
```diff
@@ -140,13 +140,64 @@ def _create_nodes_from_link(
...
+    far apart, which may indicate an input-network issue.
+    """
+    # Keep this as a local threshold for now; if users need to tune it, we can
+    # promote it to WranglerConfig (e.g., MODEL_ROADWAY) in a follow-up change.
+    DUPLICATE_NODE_WARN_SPREAD_METERS = 100.0
+    duplicate_nodes = nodes_df[nodes_df.duplicated(subset=["model_node_id"], keep=False)]
+    if len(duplicate_nodes) == 0:
+        return gpd.GeoDataFrame(nodes_df, geometry="geometry", crs=LAT_LON_CRS)
+
+    WranglerLogger.debug(
```
e-lo: i'd promote this to either info or warning.  most people aren't reading debugs

network_wrangler/roadway/links/connectors.py
e-lo: usage doesn't include all required args: oad_net: RoadwayNetwork,
    zones_table: DataFrame[ZonesTable],
    mode: str,
    local_crs: str,
    zone_buffer_distance: int,
    num_centroid_connectors: int,
    max_mode_graph_degrees: int,

network_wrangler/utils/transit.py
```diff
@@ -1072,6 +1070,12 @@ def create_connector_links_for_poor_match_stops(
     )
     WranglerLogger.debug(f"connector_links_gdf:\n{connector_links_gdf}")
 
+    # Apply default link attributes
+    if default_link_attribute_dict is None:
```
e-lo: this is duplicative of the actions a few lines below

network_wrangler/roadway/links/connectors.py
```diff
@@ -0,0 +1,414 @@
...
+    )
+    summary_df = pd.merge(left=zones_table, right=summary_df, how="left", validate="one_to_one")
+    summary_df["num_connectors"] = summary_df["num_connectors"].fillna(0).astype(int)
+    WranglerLogger.debug(f"summary_df:\n{summary_df}")
+    WranglerLogger.info(
+        f"num_connectors added per {zone_id_label} (target:{num_centroid_connectors}):\n"
+        f"{summary_df['num_connectors'].value_counts()}"
+    )
+
+    road_net.nodes_df = pd.merge(
```
e-lo: merge destroys nodes index column : model_node_id_idx.  Can you just use map instead?     # Attach the per-zone count to the centroid nodes. Use .map() rather than pd.merge()
    # so that nodes_df keeps its `model_node_id_idx` primary-key index and its attrs.
    num_connectors_col = f"{zone_id_label}_num_connectors"
    road_net.nodes_df[num_connectors_col] = (
        road_net.nodes_df["model_node_id"]
        .map(summary_df.set_index("zone_id")["num_connectors"])
        .astype("Int64")
    )
    road_net._mark_modified()

network_wrangler/roadway/links/connectors.py
```diff
@@ -0,0 +1,414 @@
...
+    WranglerLogger.debug(f"link_mode_variables:{link_mode_variables}")
+    for link_var in link_mode_variables:
+        centroid_links_df[link_var] = False
+    for link_var in MODES_TO_NETWORK_LINK_VARIABLES[mode]:
+        centroid_links_df[link_var] = True
+
+    if "highway" in road_net.links_df.columns:
+        centroid_links_df["highway"] = zone_id_label
+
+    for colname, default_value in (default_link_attribute_dict or {}).items():
```
e-lo: need a default value for LANES for centroid connectors.

e-lo (changes requested):
A few more changes to make sure things are all good! but fixes generally look awesome.

PR #482, opened 2026-07-24
chore: remove stray root-level roadway/transit/utils stub dirs

e-lo:
These incoherent re-export stub files were accidentally committed to the repo root (instead of network_wrangler/) in 6648c57 "Fix linting". They import symbols that don't exist at their paths, so they can never import successfully. They have no __init__.py, aren't on the import path, and aren't packaged, so removing them is a no-op at runtime — it just clears dead, confusing files. The real package remains under network_wrangler/.


lmz:
Nice catch; looks fine to remove!

lmz (approved):

